PR #15, opened 2026-08-27
Add first-class /plan mode with a hard read-only fence

David-T-Campos:
## Summary

Adds a first-class `/plan` workflow without adding another MCP schema or increasing the existing Core discovery budget.

- `/plan <objective>` enters a session-scoped planning phase.
- `/plan` enters planning for the current task.
- `/plan run` approves execution and lifts the mutation fence.
- `/plan clear` cancels Plan Mode.
- While planning, `read`, `view_image`, `find`, `session`, and `observe` remain available.
- `apply_patch`, `exec_command`, `write_stdin`, `agents`, and `computer` are hard-blocked by the app before their handlers execute.
- State is keyed by the durable local session, so Compact & Resume naturally carries Plan Mode to the replacement ChatGPT conversation.
- The latest directive sequence is persisted so an old `/plan` cannot re-activate after a later `/plan run` or `/plan clear` falls out of a bounded history window.
- Mutation calls fail closed when a Plan session is active but caller identity cannot be proven.

## Why no new `plan` MCP tool?

Core deliberately caps its worst-case no-query discovery at seven live schemas. Adding a ninth declared tool would increase that budget on every relevant conversation. This implementation keeps the existing surface unchanged and treats `/plan` as a session directive enforced in front of the current handlers.

## Safety model

Plan Mode is a runtime permission boundary, not merely a prompt. Even if the model ignores the planning instruction and attempts an edit, command, worker spawn, or desktop input, the connector refuses it.

`exec_command` is intentionally blocked wholesale during planning rather than trying to classify apparently read-only shell commands: shell redirection, command substitution, hooks, aliases, and program-specific options make a general shell allowlist unsafe.

## Tests

Adds focused tests for slash parsing and the read-vs-mutate permission policy. The branch also preserves the existing Core/Desktop tool declarations and discovery shape.

## Current limitation

The slash directive is derived from the app's recorded authored-user history, so Plan Mode requires session recording to be enabled. Fresh installs currently enable recording by default. A future browser-side directive path could remove that dependency without changing the enforcement layer introduced here.

### DIFF
--- a/src/main/mcp/surfaces.ts
+++ b/src/main/mcp/surfaces.ts
@@ -40,26 +40,63 @@ export interface SurfaceDefinition {
   id: SurfaceId;
   /** MCP server name. Stable; ChatGPT keys its cached metadata off it. */
   serverName: string;
-  /** Exactly what the user should type as the connector name in ChatGPT. */
+  /**
+   * Exactly what the user should type as the connector name in ChatGPT.
+   *
+   * Offered as copyable text rather than described, because the name is also the
+   * retrieval handle: `paths=["…"]` is matched against it, and a user who invents
+   * "my pc" gets a surface the model cannot address by name.
+   */
   connectorName: string;
-  /** Exactly what the user should paste as the connector description. */
+  /**
+   * Exactly what the user should paste as the connector description.
+   *
+   * This is the single most load-bearing string in the whole design. Before any
+   * discovery has happened the model holds the server name and this sentence and
+   * nothing else, and it decides from them alone whether to pull this surface's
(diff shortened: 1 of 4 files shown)


totec448-spec:
Closing this implementation after architecture review. It introduces a second plan authority inferred from recorded slash commands, while the current product has durable task/input ownership and a session-scoped plan surface. Its blanket exec/worker fence also changes existing workflows. A future explicit Plan permission feature should use the existing owners and be reviewed as its own product change; this is not a necessary bug fix.

PR #16, opened 2026-08-27
Strengthen Goal Mode completion auditing and anti-loop behavior

David-T-Campos:
## Summary

Strengthens Goal Mode's final continuation decision without changing its transport, browser delivery, idempotency, OpenRouter credential handling, or editable user prompt.

The post-transcript Goal trailers now make completion auditing explicit:

- derive the smallest completion checklist from what the user actually requested;
- keep that request as a hard scope ceiling instead of inventing polish, tests, features or follow-ups;
- distinguish conversational deliverables (where the answer itself can be the work) from computer-use/external-action tasks;
- for computer-use work, do not treat a confident `done` / `saved` / `fixed` / `built` / `tests pass` sentence as completion when the transcript still shows the relevant work as planned, promised, failed, skipped, pending, or missing verification ChatGPT itself said it would perform;
- prefer current concrete results over earlier plans or claims;
- continue with one targeted next message naming the exact unmet condition rather than a generic `continue`;
- answer an assistant question when the answer is already unambiguous from the conversation;
- add a repeated-blocker escape hatch: after the same blocker survives three consecutive assistant turns and genuinely requires user-only information, permission, or credentials, stop instead of generating an infinite continuation loop.

## Why the trailer?

Goal's editable policy prompt appears before a potentially very long transcript. The existing architecture already appends a short app-owned trailer after the transcript specifically because the last instruction is more reliable in long contexts. This PR puts the completion contract at that load-bearing boundary rather than replacing user-customized prompts or changing the wire protocol.

## Scope

This PR intentionally does **not** add `/plan`, local-tool evidence export, new MCP schemas, or browser automation changes. It is independent of the separate Plan Mode PR.

## Tests

Adds focused contract tests covering:

- computer-use completion auditing;
- scope-ceiling behavior;
- the three-turn repeated-blocker escape hatch;
- preservation of the existing `NO_REPLY` stop protocol.

## Follow-up

A stronger future layer could feed Goal a privacy-bounded app-authored evidence ledger (tool outcome + changed-path metadata, excluding raw command output/file contents). This PR is deliberately the smaller, reviewable policy improvement first.


totec448-spec:
Closing this implementation after review. The prompt changes impose a new completion policy on Goal/Loop, including treating blocked work as NO_REPLY, without repairing a demonstrated ownership or completion defect. This can suppress a needed user-visible blocker. Completion must remain tied to the existing Goal/finish contract and the user objective.

PR #18, opened 2026-08-27
feat: add Claude-style /loop mode

David-T-Campos:
## Summary

Adds a first-class `/loop` / `/proactive` workflow modeled on the current Claude Code loop contract, implemented clean-room for Chat On Steroids.

Research was cross-checked against current Claude Code scheduled-task documentation, current extracted prompt/tool fragments tracked for Claude Code v2.1.246 (August 25, 2026), and publicly reconstructed runtime code. The key finding is that modern `/loop` is not one prompt or one scheduler: fixed cadence, dynamic self-pacing, autonomous maintenance, monitoring/fallback wakes, and scheduler lifecycle are separate cooperating layers.

## Behavior

- `/loop 5m <task>` (compact `s/m/h/d`) starts a fixed loop.
- `<task> every 5 minutes` is recognized as the equivalent fixed form.
- `/loop <task>` starts a dynamic/self-paced loop.
- `/loop 15m` and `/loop every 15 minutes` run autonomous maintenance at that fixed cadence.
- bare `/loop` runs autonomous maintenance dynamically.
- maintenance prefers project `.claude/loop.md`, then `~/.claude/loop.md` when accessible, and re-reads the selected file on each iteration so edits take effect.
- `/proactive` is accepted as an alias.
- The requested task runs once immediately; scheduling controls subsequent iterations.
- Fixed intervals have a one-minute minimum/granularity.
- Dynamic iterations choose their next wake between 60 and 3600 seconds, or explicitly stop themselves.
- A bounded ~20-minute fallback heartbeat prevents a dynamic loop from silently dying if an iteration forgets to schedule the next wake; if the fallback iteration also fails to re-arm, the accidental poller retires.
- Dynamic loops stop at the seven-day lifetime without an extra wake. A fixed recurring loop that reaches the seven-day age limit is allowed its next due fire once, then is deleted, matching Claude's recurring-task age-out behavior.
- If a fixed fire is missed while the conversation is busy, missed periods collapse into one pending iteration rather than a catch-up burst.
- Loop state is durable across app/browser restart and remains scoped to the proven calling conversation/session.
- Escape clears waiting dynamic loops without cancelling fixed loops.

## Architecture

- New `src/main/loop.ts` owns parsing, durable task state, due-fire selection, final-age-out handling, delivery leases, wake/stop/cancel operations and expiry.
- Extends the existing Core `session` tool rather than adding another permanent MCP tool/schema.
- Recorder evidence remains authoritative for the `/loop` directive; model-supplied input is only a consistency check.
- Bridge + extension delivery reuse the existing conversation/tab ownership model and acknowledged-send path.
- Protocol version is bumped because the bridge surface changes.
- No raw filesystem/tool payloads are persisted by the loop scheduler.

## Claude runtime model established by research

Current Claude Code composes several mechanisms:

1. The bundled `/loop` Skill parses the request and executes the requested work immediately once.
2. Explicit cadence uses `CronCreate`; future due prompts are queued and delivered between turns rather than interrupting an active turn.
3. Prompt-only loops self-pace. The model performs an iteration and then uses `ScheduleWakeup` for its next 1–60 minute wake or explicitly stops. Current dynamic orchestration may prefer a persistent `Monitor` when there is a concrete event to wait for and retains a scheduled fallback heartbeat.
4. Bare/interval-only loops enter separate autonomous-maintenance behavior, with project/user `loop.md` support and conversation/PR work prioritized over invented busywork.
5. Scheduler runtime owns busy deferral, no-catch-up behavior, task limits, local-time cron handling, expiry, and cron jitter.

## Deliberate platform adaptations

This ports the behavior onto Chat On Steroids rather than pretending Anthropic-only infrastructure exists:

- Claude fixed loops are cron-backed and can round awkward intervals such as `7m`/`90m` to a clean cron-representable cadence, with deterministic cron jitter. The local COS scheduler can represent those intervals directly, so it preserves the requested minute cadence instead of inventing Claude's model-chosen rounding decision or jitter.
- Claude dynamic wake targets are ultimately minute-rounded by its cron-backed wake implementation and may be adjusted around prompt-cache TTL. COS uses the model-selected 60–3600 second local wake directly because it has no equivalent prompt-cache scheduler.
- Claude may use native persistent `Monitor` tasks; COS has no equivalent Anthropic service, so dynamic loops use the local wake/fallback mechanism.
- Current `ScheduleWakeup` also exposes presentation/control details such as `noop` collapsing and `stop`; COS exposes explicit local `loop_wakeup` and `loop_stop` operations rather than fabricating terminal/push-notification semantics it does not own.
- Claude's autonomous-loop machinery can inject cached/truncated `loop.md` content (25,000-byte cap) into turns. COS instead instructs the model to re-read the accessible file each iteration, which keeps edits live without duplicating Claude's prompt-cache internals.
- Claude can offer cloud schedules/push behavior in supported environments. This PR stays local and does not emulate those proprietary services.
- Claude local loops are session-scoped and restored via resume/continue. COS persists the loop state in its existing durable app state, so browser/app restart recovery is native to this harness.

## Safety / idempotency

- Loop control requires a proven ChatGPT caller and unique recorded session.
- The same recorded `/loop` user turn cannot create duplicate tasks.
- Delivery uses expiring leases tied to the claiming browser client.
- A final aged fixed tick is deleted only after successful browser delivery; a failed send remains retryable.
- Unknown/stale loop IDs fail rather than mutating another task.
- Per-session task count, prompt size, wake delay and lifetime are bounded.

## Validation

The exact final feature tree is:

`df38c02c03988b9a0ce4b9d03b1d364fc80aa3a3`

It was validated on GitHub Actions before squashing with:

- `npm ci`
- `npm run rg`
- `npm run typecheck`
- `npx vitest run --exclude test/mcp-shutdown.test.ts`
- `npx vitest run test/mcp-shutdown.test.ts`
- `npm run build`

Final results:

- **1,732 tests passed, 93 skipped, 0 failed** across the main suite.
- **2/2 shutdown tests passed** in isolation.
- Production Electron/Vite build passed for main, preload and renderer.
- The Core MCP discovery-size budget remains enforced by the test suite.

`npm run verify:ci` was also attempted first, but its `verify:privacy` step audits the fork's entire inherited Git history and stops on historical upstream commits containing old maintainer email addresses. Those commits predate this PR and are not changed by it; the actual typecheck, complete code test suite, shutdown isolation and production build were therefore run explicitly as listed above and all passed.

The branch was then squashed by reusing the exact byte-tested Git tree. Final PR head:

`0f31b56496bb7ba3628fc934f0143f51de9916a3`

## Notes

This is a clean-room behavioral implementation based on official public documentation plus publicly shipped/extracted/reverse-engineered artifacts used to establish observable mechanics. It does not copy or redistribute Anthropic proprietary source code or long internal prompts verbatim. A reconstructed source hint for `/loop until <condition>` was found, but no sufficiently proven current execution semantics were established in the official docs/runtime path, so this PR deliberately does **not** invent that feature.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -24,7 +24,7 @@ const PORTS = [8765, 8766, 8767, 8768, 8769];
 const HELLO_TIMEOUT_MS = 1200;
 const REQUEST_TIMEOUT_MS = 10_000;
 /** Bumped only when the request/response shape changes; the app compares it. */
-const BRIDGE_PROTOCOL = 8;
+const BRIDGE_PROTOCOL = 9;
 
 /**
  * Journal caps. The byte figure is what actually matters — chrome.storage.session has a
@@ -1947,6 +1947,40 @@ const HANDLERS = {
     });
     return ownsDocument(source) ? result : { ok: false, error: 'stale_document' };
   },
+  /** Claim one due /loop prompt only for the document that owns this exact chat. */
+  async loop_due(message, _sender, source) {
+    await load();
+    if (!ownsDocument(source)) return { ok: false, error: 'stale_document' };
+    const conversationId = cleanConversationId(message.conversationId);
+    if (!conversationId) return { ok: false, status: 400, error: 'bad_conversation_id' };
+    await noteTabConversation(source, conversationId);
+    if (!ownsDocument(source)) return { ok: false, error: 'stale_document' };
+    if (!(await deliverConversationJournal(conversationId))) {
+      return { ok: false, status: 503, error: 'transcript_not_delivered', retryable: true };
+    }
+    const result = await call('/loop/due', {
+      method: 'POST',
+      body: JSON.stringify({ conversationId, clientId: String(source.tab) })
+    });
+    return ownsDocument(source) ? result : { ok: false, error: 'stale_document' };
+  },
+  async loop_ack(message, _sender, source) {
+    await load();
+    if (!ownsDocument(source)) return { ok: false, error: 'stale_document' };
+    const conversationId = cleanConversationId(message.conversationId);
+    if (!conversationId) return { ok: false, status: 400, error: 'bad_conversation_id' };
+    return call('/loop/ack', {
+      method: 'POST',
+      body: JSON.stringify({ conversationId, token: String(message.token || ''), sent: message.sent === true })
+    });
+  },
+  async loop_escape(message, _sender, source) {
+    await load();
+    if (!ownsDocument(source)) return { ok: false, error: 'stale_document' };
+    const conversationId = cleanConversationId(message.conversationId);
+    if (!conversationId) return { ok: false, status: 400, error: 'bad_conversation_id' };
+    return call('/loop/escape', { method: 'POST', body: JSON.stringify({ conversationId }) });
+  },
   /**
    * The goal loop: this page saw its turn genuinely finish and wants the next user message.
    *
@@ -2201,6 +2235,9 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
     'closed',
     'compact',
     'auto_compact_claim',
+    'loop_due',
+    'loop_ack',
+    'loop_escape',
     'goal_draft',
     'goal_focus',
     'goal_ack',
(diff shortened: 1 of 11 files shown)


totec448-spec:
Closing this implementation as superseded by the current Goal/Loop architecture. It adds a separate durable loop scheduler and additional session actions alongside the existing Goal driver, input outbox and continuation ownership. A concrete missing Loop behavior should be fixed in those owners rather than introducing a second scheduler.

Issue #21, opened 2026-08-28
Auto-compaction misses an oversized active session, then compacts its own handoff after the 10-minute TTL


johmaru:
### App version

v2.0.2

### Operating system and version

Windows 11 Pro 25H2

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT workspace / custom MCP access

Pro (custom MCP read/fetch only)

### What happened

Auto-compaction was enabled with a threshold of 400,000 tokens, but it did not trigger even after the session exceeded that threshold and continued through long-running turns and tool calls.

The session eventually grew to approximately 730,000 context tokens while work was still actively continuing. Since automatic compaction still had not occurred, I manually started Compact & Resume at approximately 21:59.

I was using Thinking effort: Pro at the time.

ChatGPT began generating the handoff brief normally, but the handoff generation continued for longer than 10 minutes.

Around the 10-minute mark, the behavior unexpectedly reversed: auto-compaction appears to have triggered against the handoff generation itself.

The in-progress handoff was stopped, a new Compact & Resume continuation was created, and the same handoff prompt was submitted again.

So I observed two related problems:

Auto-compaction did not trigger when it should have.
The session was already well above the configured 400k threshold and continued through long-running turns and tool calls, eventually reaching approximately 730k context tokens. I had to start Compact & Resume manually.
Auto-compaction later triggered when it should not have.
After the manually started handoff had been generating for approximately 10 minutes, the handoff generation itself appears to have been treated as an eligible turn for auto-compaction. It was stopped and another Compact & Resume attempt was started.

I expected auto-compaction to handle the original oversized active session. Once a Compact & Resume handoff generation had started, I expected that generation to be excluded from automatic compaction until it either completed or failed cleanly.

Instead, auto-compaction did not act on the original oversized workload, but later appears to have acted on the compaction operation itself.

No handoff was successfully stored, and the original session remained at approximately 730k context tokens.

### Steps to reproduce

Enable automatic compaction.
Set the automatic compaction threshold to 400,000 tokens.
Continue working in the conversation after it has exceeded the threshold, including long-running turns and tool calls.
Observe that automatic compaction does not trigger.
In my case, the session continued growing until it reached approximately 730,652 context tokens.
Set Thinking effort to Pro.
Start Compact & Resume manually.
Allow the handoff generation to continue for longer than approximately 10 minutes.
[...]

PR #39, opened 2026-08-30
Keep a running handoff generation out of auto-compaction

Maximapple:
Refs #21 — the second half of that report.

## Root cause

A Compact & Resume handoff is one ChatGPT generation, and on a ~730k context it can legitimately run longer than `CONTINUATION_TTL_MS`. The continuation lease was measured from `openedAt` alone:

```ts
const isOpen = (entry) => … && Date.now() - entry.openedAt < CONTINUATION_TTL_MS;
```

So the sweep gave the transaction up while the browser was still generating the brief. Once it did, the app stopped projecting a busy job for that chat and went back to reporting it as eligible. The page — which was still capturing that same generation — read "eligible", interrupted the turn, and started a second Compact & Resume against its own handoff. That matches the reported timeline exactly: the reversal at about the ten-minute mark, the in-progress handoff stopped, a new continuation created, and the same prompt resubmitted with no handoff stored.

## Change

Two independent halves; either alone breaks the loop.

**The lease renews.** A continuation carries `touchedAt` alongside `openedAt`, and `/activity` accepts a `compactToken` that renews it. `touchContinuation` accepts only an exact armed capture in `awaiting-summary` for the same session *and* the same chat, and refuses a token that has already been silent for a full TTL — a lapsed transaction cannot resurrect itself. Renewals are persisted at most twice a minute, so a busy poll loop does not turn into a durable write loop while the last written value stays comfortably inside the expiry window. `restoreContinuations` prefers `touchedAt` and falls back to `openedAt`, so snapshots written before this field restore exactly as they do today.

**The page refuses regardless.** `maybeAutoCompact` now treats its own `compactCapture` as work that owns the chat, so whatever the app reports, a page cannot select the generation it is itself capturing.

`background.js` forwards the token verbatim without inspecting it; only the app can decide whether it names an armed capture.

## Validation

`test/continuation.test.ts`, four cases on the real transaction:

- an actively reported handoff stays open past the fixed TTL (two renewals at 66% of the window each, which without renewal is past expiry);
- a full window of silence still expires it, the sweep aborts it, and the lapsed token is refused;
- only the exact armed generation is renewed — wrong chat, wrong session, unknown token, an unarmed continuation, and one already past capture are all refused;
- a snapshot restores against the renewed lease rather than the original open time.

`test/content-script.test.ts`, one behavioural case reproducing the report from the browser side: a compaction starts, the app then goes back to reporting the chat as eligible (the state the lapsed lease produced), and the page must still start exactly one compaction. It also asserts the capture token appears on the polls after the capture and not before it.

Full local run on Windows x64: `npm run typecheck` clean, `vitest run` 1809 passed / 18 skipped across 68 files, `test/mcp-shutdown.test.ts` 2 passed.

## Not included

The first half of #21 — auto-compaction not triggering on the oversized session in the first place, before any handoff existed — is a different path and is deliberately untouched here. This PR only stops the compaction from eating its own handoff; the issue should stay open for that half.

### DIFF
--- a/src/main/bridge.ts
+++ b/src/main/bridge.ts
@@ -127,6 +127,7 @@ import {
   CONTINUATION_TTL_MS,
   continuationByToken,
   continuationForSession,
+  touchContinuationForChat,
   pendingAutomaticContinuations,
   supersededSourceConversations,
   dispatchContinuationDestinationSendNow,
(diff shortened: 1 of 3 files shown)


totec448-spec:

The handoff-lifetime fix is incorporated through an exact-answer progress adaptation: bounded progress renews the manual continuation deadline, while unrelated activity cannot keep it alive. Closing this source PR does not close the separate initial auto-compaction report in #21.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/session/continuation.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/continuation.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

PR #43, opened 2026-08-31
RFC: define a bounded Codex Desktop thread bridge

hhh2210:
## What changed

This remains a design-only RFC for a bounded bridge from Chat On Steroids Core to Codex Desktop threads, but one major assumption has changed since the draft was opened.

Codex CLI 0.152.1 now ships an official:

```text
codex queue --thread <UUID> --message ...
```

On a Mac running Codex Desktop, I tested that command against the UUID of a task already visible in Desktop. The queued message became the next user turn in that exact task and the task continued there, without foreground activation or GUI automation.

That removes the original “can we background-send to the actual Desktop task at all?” blocker. The remaining problem is narrower: verify the supported `list` / `read` / `wait` surface and its thread identity / stale-thread guarantees, then wrap only the bounded subset CoS needs.

The RFC in [`docs/codex-desktop-bridge.md`](docs/codex-desktop-bridge.md) has been updated around that evidence.

## Proposed first slice

One model-facing `codex_desktop` schema, initially focused on existing threads:

- `list`
- `read`
- `send`

with exact thread-id addressing, separate read/control grants, bounded results, and read-back after mutation when the supported Codex surface permits it.

`wait` and `create` can follow if their supported contracts are equally clear; there is no reason to invent a private Desktop pipe merely to obtain send.

## Why keep this as a draft

The transport is no longer hypothetical, but I still do not want to land runtime code before two product decisions are clear:

1. Is supervising Codex Desktop threads in scope for CoS Core, or better kept as an external/meta-harness integration?
2. Is wrapping the shipped exact-thread CLI/app-server surface an acceptable dependency, once its read/list identity guarantees are verified?

## Safety / ownership constraints

- no foreground activation of Codex Desktop;
- no title-only mutation;
- no direct database writes;
- no private-pipe bypass;
- no generic model-controlled RPC passthrough;
- no silent fallback to a different thread/daemon after failure;
- thread reads remain bounded and privacy-reduced;
- browser work stays inside the target Codex Desktop task, which can use its own `control-chrome` state.

## Current evidence

- [x] exact-thread background send reaches the same task already visible in Codex Desktop;
- [x] no GUI focus/click/type is required for that send;
- [x] RFC updated to remove the obsolete transport blocker;
- [ ] supported enumeration/read/status semantics verified;
- [ ] CoS runtime implementation added;
- [ ] live `list → read → send → read-back` acceptance through CoS.

## Review questions

1. Is this integration in scope for the project?
2. Does the one-schema / separate read-control grant model fit the Core surface budget?
3. Should the first implementation stop at existing-thread `list/read/send`?
4. Once verified, is the first-party CLI/app-server thread surface an acceptable dependency, or would you prefer another supported Codex contract?

### DIFF
--- a/docs/codex-desktop-bridge.md
+++ b/docs/codex-desktop-bridge.md
@@ -4,37 +4,42 @@
 
 Design draft only. This document deliberately adds no production tool yet.
 
-The missing piece is not the thread API. Codex app-server already has bounded methods for
-listing, reading, creating and messaging threads. The missing piece is a supported way for a
-separate local app to bind those methods to the **specific Codex Desktop instance the user is
-looking at**, rather than to an unrelated CLI/default daemon.
+The transport picture changed after this RFC was opened. Codex CLI 0.152.1 ships an official
+`codex queue --thread <UUID> --message ...` command against the shared local app-server. On a Mac
+running Codex Desktop, queuing to the UUID of a task already visible in Desktop delivered the next
+user turn to that exact task and the task continued there, without foreground activation or GUI
+automation.
 
-Implementing before that ownership boundary is agreed would recreate the bug this proposal is
-meant to remove: a successful backend acknowledgement that affected the wrong task.
+That materially shrinks the original blocker: background **send to a known existing Desktop thread**
+is now proven through a supported Codex surface. The remaining integration question is narrower:
+which supported list/read/wait primitives provide bounded identity and transcript/status evidence
+for those same Desktop threads, and what host-instance guarantees should Chat On Steroids require
+before it exposes them to the model.
 
 ## Problem
 
-Chat On Steroids can inspect local recordings and can control the desktop, but neither is the
-right default for supervising an existing Codex Desktop task:
+Chat On Steroids can inspect local recordings and can control the desktop, but neither is the right
+default for supervising an existing Codex Desktop task:
 
 - GUI automation steals focus and competes with the user's foreground work;
 - writing into Codex's persisted databases bypasses its ordering and notification contracts;
-- invoking `codex queue` or another CLI command may reach a different app-server daemon;
-- a successful CLI response therefore does not prove the visible Desktop task received anything.
+- title-only or default-daemon addressing is ambiguous;
+- a transport acknowledgement alone is not enough unless the bridge can name and read back the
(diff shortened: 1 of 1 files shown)


winnal:
yeah, i've actually been exploring a similar direction, basically having CoS (Codex on Steroids) being able to directly message or prompt Codex desktop threads, or create new ones. This gives best of both worlds, access to using Codex Desktop to spend the usage limits we have available, like using Luna Max, while the CoS uses Sol High. I was exploring other plugins that target this scope of functionality, but it does not have the full fledged functionality of CoS on ChatGPT web. So would be nice to have both and integration with Codex Desktop. I was exploring for example having CoS sending a message to a Codex Desktop thread so that it could wake it upon CoS finishing its turn, so that it can create a loop where Codex Desktop can check CoS's work, or vice versa. This could be an alternative way of the goal feature currently requiring OpenRouter API key for a secondary model to write the continuation prompt. Currently, I'm actually working on my own integration with CoS that bridges between ChatGPT web and Codex Desktop. But it's more of a 3rd party meta-harness on top of Codex App Server + CoS that basically allows CoS to orchestrate workers not just as other chat tabs, but Codex threads. While on the other hand, I also have a dedicated "prime" agent similar to CoS, but on Codex Desktop, which monitors and checks CoS prime agent whenever it finishes its turn. So both ways works.

hhh2210:
Concrete update since this RFC was opened: Codex CLI 0.152.1 now ships an official `codex queue --thread <UUID> --message ...` command against the shared local app-server. I tested it against the exact thread that was already visible in Codex Desktop: the queued message appeared as the next user turn in that same thread and the Desktop task continued from it, without foreground activation or GUI automation.

That removes the biggest transport uncertainty in this draft. The likely first slice can now be much smaller: wrap supported Codex thread identity/list/read/send primitives, keep CoS responsible for permissioning + acknowledgement/read-back, and avoid inventing a private socket transport. The remaining thing I want to verify is the supported enumeration/read surface and host identity guarantees, not whether background send is possible at all.

@winnal this is essentially the wake/loop direction you described. @totec448-spec, if this still fits the project scope, I can update the RFC around the now-real official transport and reduce the proposed implementation surface substantially.

PR #46, opened 2026-09-01
2.0.3 rebuild: Loop, per-chat goal mode, durable continuation tickets

totec448-spec:
The 2.0.3 line rebuilt from 2.0.2. **Not for merge or release yet** — filed so CI runs it
on all three platforms and so the diff is reviewable.

## What is in it

**Loop.** Goal stops when the second model decides the job is done, and on a long unattended
run that judgement is the part that fails. Loop is the same machinery with the stop removed:
every finished turn gets a message back, and only the Loop switch ending it ends the run.
Goal and Loop are two positions of one stored mode rather than two booleans, so no page,
poll or settings window has a state where both are on. "Never no reply" is held at three
layers — a third editable prompt, a response schema whose action enum contains only
`continue`, and bounded regeneration that ends in a retryable `loop_stop_refused` rather than
a sentence this app wrote and attributed to the model.

**A goal's mode is pinned to its chat.** A goal written from the New Chat sheet starts the
chat immediately, but the mode it then ran in was read from the standing switch — off,
therefore Goal — so a Loop written and launched in one press could end at the second turn on
one "looks done". The mode is now pinned as that chat's own switch in the same request,
before the goal exists to be acted on, and clearing the goal turns it back off.

**The continuation is a ticket.** It is opened durably *before* any fallible page barrier
rather than after it, so a composer error before Send leaves the obligation standing. Auto
Off cancels only the tickets a threshold created; manual Compact & Resume stays explicit.
Three fifteen-minute checkpoints reload the same exact chat rather than failing it, and
exhausting them ends only the automatic browser action — the continuation stays collectable.

**Recovery and attribution.** Single-owner tunnel restarts, hidden activity delivery kept
live, unread exec results preserved under capacity, request ids pinned to their chat, reload
loops stopped, Project conversation routes recognised wherever a chat is named, activity
treated as a grant so a chat that stops earning it is reloaded, and bounded agent-tab
recovery.

**Fresh installs start with `allowUnattributedCalls` on.** The ambiguity fences refuse work
when the extension cannot prove the caller, and a new install is where that evidence path is
least likely to be healthy yet — so the first thing a new user saw was a refusal. The work
now runs and its activity stays labelled Unattributed rather than being guessed onto a chat.
`DEFAULT_MULTI_AGENT` keeps `false`, so an upgrade never relaxes an older config over a field
that config never wrote.

**The privacy gate exempts already-published history.** Two merge commits on `origin/main`
carry a non-noreply maintainer address — the forge wrote them when #19 and #20 were merged,
and no local hook was ever in a position to see them. They are public already, so failing
every later local push cannot unpublish them; it only stranded the working clone with no
clean commit available that would clear the check. Commits reachable from `origin/main` are
now exempt. Everything a push would actually add stays checked, and a missing `origin/main`
exempts nothing.

## Review notes

- **This PR conflicts with `main` and is not mergeable as it stands.** The branch is 6
  commits behind, and three of those are the forge's merges of work this branch already
  carries under different SHAs (`Scope the public-history privacy gate`, `Recognise Project
  conversation routes`, `preserve redeemed resume lease past ack deadline`). The conflicting
  paths are `extension/chatgpt-dom.js`, `scripts/verify-public-history.mjs`,
  `src/main/bridge.ts` and `test/public-history-privacy.test.ts`. Bringing the branch up to
  date is deliberately left out of this PR.
- Because the PR does not merge cleanly, GitHub has no `refs/pull/46/merge` for the
  `pull_request` job to check out, so no CI ran on the PR event. CI was instead dispatched
  directly against the branch head — see below.
- `package.json` is still at `2.0.2`; no version bump or release notes are in this PR.

## Verification

`npm run verify` locally on Windows x64: privacy gate passed (40 commits, 3 tags), typecheck
clean, Vitest 71 files / 2019 passed / 18 skipped including the isolated
`mcp-shutdown` run. `npm run dist:x64` builds, and the resulting NSIS installer installs and
launches. macOS and Linux are on CI here rather than locally.


totec448-spec:
Superseded by a fresh PR from the same branch, now merged with main and green locally.

Issue #22, opened 2026-08-28
Firefox support


BookCatKid:
### Problem or workflow

Please add firefox support	

### Proposed change

should be pretty easy, just change a few apis idk

Maximapple:
There is already unlanded work for this that nobody seems to have noticed, so worth connecting the two.

[#47](https://github.com/totec448-spec/chat-on-steroids/pull/47) (@nabeel2k11, opened for a different bug) carries a `scripts/stage-firefox-extension.mjs` with a `firefoxManifest()` that generates the Gecko MV3 delta from the canonical manifest — swapping `background.service_worker` for `background.scripts`, adding the `extension_pages` CSP, dropping `minimum_chrome_version` — plus `test/firefox-extension-package.test.ts` so the two manifests cannot drift, and a release-workflow step to stage the package.

None of it is on `main`. The manifest there is still Chromium-only:

```json
"background": { "service_worker": "background.js", "type": "module" },
"minimum_chrome_version": 116
```

The bug fix that PR was actually opened for shipped separately in 2.0.7, so the PR now reads as stale and is likely to be closed with the Firefox work still inside it. Extracting that part into its own PR would give this issue a concrete starting point — it is purely additive, so it should not carry the conflicts the rest of #47 now has.

That covers packaging, not the runtime question: the extension is one codebase for both browsers, so what remains unproven is whether `background.js` and the content scripts behave under Gecko's MV3 event pages, which suspend differently from a Chromium service worker. Worth treating as the second half of this issue rather than assuming packaging is the whole of it.

totec448-spec:
hey @BookCatKid, Firefox remains a separate support request. A manifest/package conversion is useful groundwork, but it does not verify the companion’s runtime, background lifecycle or browser APIs on Firefox. Keeping this open without claiming Firefox support in the 2.1.0 bug pass. Thanks for asking.

PR #47, opened 2026-09-01
Fix stale Firefox revival tabs after restart

nabeel2k11:
## Summary

Fixes a browser-restart failure where a persisted deferred worker-revival marker could outlive its app-side command and recreate an old ChatGPT conversation on every Firefox startup.

### Root cause

`deferredRevivals` intentionally survives browser restarts, but 2.0.2 treated that browser-local marker as sufficient authority to recreate `/c/<conversation>?clf=<command>`. The desktop app can meanwhile commit, cancel, supersede or retire the command. A stale marker therefore remained capable of opening UI even when the app's durable command queue was empty.

### Fix

- add authenticated read-only `POST /commands/revivals/pending` batch validation
- reconcile persisted revival markers against app-owned command truth before any recovery `tabs.create`
- durably prune stale markers
- fail closed on bridge/version/transport errors: keep the inert marker for later retry, but open no tab
- keep validation read-only so a confirmed live wake can still be redeemed by the exact worker page afterwards
- bump bridge protocol to 9 so mixed old/new peers cannot silently fall back to the unsafe behavior

### Firefox release path

- bump app/extension release metadata to 2.0.3
- add deterministic Firefox staging from the shared extension source
- preserve the existing AMO add-on id `chat-on-steroids-companion@local`, Firefox 128 minimum and required data-collection declarations
- add `Chat-On-Steroids-Firefox.zip` to release candidate/checksum/publication workflows
- add 2.0.3 changelog/release notes

## Validation

- TypeScript: pass
- production `electron-vite build`: pass
- main Vitest matrix: **1815 passed, 18 skipped, 0 failed**
- shutdown suite: **2 passed, 0 failed**
- focused extension suite: **83/83 passed**
- focused bridge suite: **130/130 passed**
- `git diff --check`: pass
- staged privacy check: pass
- generated Firefox manifest/package verified locally

`npm run verify:ci` currently stops at the public-history privacy gate because two commits already reachable from upstream `main` use a non-noreply maintainer identity (`9e27c0fafc20bf2c81509844d5f92868678b4168` and `03acfbaad9d753d09487761e97cc1eade8eb8b22`). That ancestry issue predates this change; this commit itself passes `verify-public-history.mjs --staged`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -86,6 +86,7 @@ jobs:
             'Chat-On-Steroids-Linux-arm64.AppImage',
             'Chat-On-Steroids-Linux-arm64.deb',
             'Chat-On-Steroids-Extension.zip',
+            'Chat-On-Steroids-Firefox.zip',
             'SHA256SUMS.txt'
           ]) {
             if (!notes.includes(`\`${artifact}\``)) throw new Error(`${notesPath} is missing ${artifact}`);
@@ -181,6 +182,7 @@ jobs:
             publish/Chat-On-Steroids-Linux-arm64.AppImage \
             publish/Chat-On-Steroids-Linux-arm64.deb \
             publish/Chat-On-Steroids-Extension.zip \
+            publish/Chat-On-Steroids-Firefox.zip \
             publish/SHA256SUMS.txt \
             --title "Chat On Steroids ${TAG#v}" \
             --notes-file "docs/release-notes/$TAG.md"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -324,6 +324,27 @@ jobs:
             unzip -Z1 release/Chat-On-Steroids-Extension.zip | grep -Fxq "$required" || { echo "Extension zip missing $required" >&2; exit 1; }
           done
 
+      - name: Package Firefox AMO extension
+        run: |
+          set -euo pipefail
+          npm run extension:firefox:stage
+          (cd release/firefox-extension && zip -qr ../Chat-On-Steroids-Firefox.zip . -x '*.map')
+          node <<'NODE'
+          const fs = require('node:fs');
+          const manifest = JSON.parse(fs.readFileSync('release/firefox-extension/manifest.json', 'utf8'));
+          const pkg = require('./package.json');
+          if (manifest.version !== pkg.version) throw new Error('Firefox extension version mismatch');
+          if (manifest.background?.scripts?.[0] !== 'background.js' || manifest.background?.service_worker) {
+            throw new Error('Firefox background manifest was not transformed');
+          }
+          if (manifest.browser_specific_settings?.gecko?.id !== 'chat-on-steroids-companion@local') {
+            throw new Error('Firefox add-on id changed');
+          }
+          NODE
+          for required in manifest.json background.js chatgpt-dom.js content.js fiber.js overlay.css popup.html popup.css popup.js icons/icon16.png icons/icon32.png icons/icon48.png icons/icon128.png LICENSE; do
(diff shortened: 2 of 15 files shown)


nabeel2k11:
Firefox package validation: web-ext lint reports **0 errors, 0 notices, 4 warnings**. Two warnings are compatibility notices for data_collection_permissions with the existing Firefox 128 floor (the key is supported from Firefox 140 desktop / 142 Android); two are pre-existing dynamic innerHTML warnings in current extension code. I left the release floor unchanged rather than reducing compatibility solely to silence lint. A local web-ext build of the generated AMO stage also succeeds.

totec448-spec:
thank you very much, 203 releases later today and it conflicts a little bit, will look at it tomorrow when release is done

Maximapple:
The bug this fixes still appears to be live on `main`, but this branch has drifted a long way
from it — flagging both rather than letting the PR sit ambiguous.

**Still unfixed:** there is no `/commands/revivals/pending` route in `src/main/bridge.ts` and no
reconciliation in `extension/background.js` on `main`. `deferredRevivals` still survives a browser
restart and is still treated as sufficient authority to recreate `/c/<conversation>?clf=<command>`,
which is exactly the root cause described here. So the reasoning holds.

**But the branch cannot be rebased as-is.** It bumps `BRIDGE_PROTOCOL` to 9; `main` is on 12. It
also carries a 2.0.3 release path — version bumps, `docs/release-notes/v2.0.3.md`, the publish and
release workflows, Firefox staging scripts and an AMO id — which is all historical now. Those
parts will conflict comprehensively and are not what the fix needs.

What is worth keeping is small and self-contained: the authenticated read-only batch validation,
the reconcile-before-`tabs.create` step, durable pruning of stale markers, and failing closed by
keeping an inert marker rather than opening a tab. Lifted onto current `main` against protocol 12
that is a much smaller diff than this one, and it would not need a protocol bump at all if the new
route is additive and the extension treats an unknown route as "do not open".

Worth checking with the maintainer whether the Firefox release path is still wanted before
redoing it — #22 is open and unclaimed, so that half may be a separate conversation now.

Maximapple:
Half of this has shipped, and the half that has not is the more interesting one.

**The bug fix landed in 2.0.7** (`b6cdb61`), without this PR being merged. Current `main` has both ends of the reconciliation:

- `src/main/bridge.ts`: `if (route === '/commands/revivals/pending' && req.method === 'POST')`
- `extension/background.js`: `const result = await call('/commands/revivals/pending', …)`

So a persisted revival marker is now validated against app-owned command truth before any recovery `tabs.create`, which is exactly the fix described here. Nothing to rebase for that part.

**The Firefox packaging half did not land.** On `main` there is no `scripts/stage-firefox-extension.mjs`, no `test/firefox-extension-package.test.ts`, and the manifest still carries only the Chromium shape:

```json
"background": { "service_worker": "background.js", "type": "module" },
"minimum_chrome_version": 116
```

Your `firefoxManifest()` is the piece that is missing — generating the Gecko MV3 delta (`background.scripts`, the `extension_pages` CSP, dropping `minimum_chrome_version`) from the canonical manifest rather than maintaining a second copy, with a test so the two cannot drift.

That is groundwork for #22, which is open with no comments and no other work attached to it. If you are willing, splitting it out as its own PR against `main` would be worth far more than rebasing this one: it is additive (a new script, a new test, a workflow step), so it should not conflict the way the revival changes now would against code that already landed.

Either way this PR is closeable in its current form — the fix it was opened for is in the product.

totec448-spec:
Closing as superseded. Current main already validates deferred revivals before startup, persists revival custody, and does not reopen a user-closed revival tab after restart; the corresponding extension tests passed in this audit. This older branch also includes Firefox packaging and version/protocol changes that should not be merged as a revival fix. Firefox support remains tracked separately in #22.

PR #58, opened 2026-09-01
Make recovery crash boundaries deterministic

### LINKED ISSUE #52: Make crash-boundary recovery tests deterministic
## Problem

Several crash/recovery tests model an important durability boundary by manually creating promises, yielding arbitrary microtasks, or waiting for a timer/poll to observe the next state.

That makes the tests sensitive to scheduler and filesystem load even when the product ordering is correct. One concrete example is the dropped-worker transport test: it must prove that the old browser command remains durable until the failed broker state fsyncs, then disappears only after the same critical persistence flight drains.

## Desired behavior

Use a tiny deterministic fault gate for recovery tests:

- production work under test calls/awaits the gate at the exact boundary;
- the test awaits proof that the boundary was reached;
- the test inspects crash-time state;
- release is explicit and idempotent;
- where a real durability barrier already exists (`persistCriticalSwarmNow`, etc.), await that barrier instead of guessing how many microtasks/timers are needed.

No production behavior should change. The goal is stronger regression evidence and less load-dependent CI flakiness.

Inmerson:
## Root cause

Crash/recovery tests were proving some durability boundaries with hand-written promises, arbitrary microtask yields, or polling. Under a loaded full suite, that can fail even when the product ordering is correct because the test is guessing when a persistence flight has drained.

## Change

- add a tiny reusable `faultGate()` test helper with explicit `reached`, `hold()` and idempotent `release()` semantics;
- move worker-binding and continuation crash barriers onto that deterministic helper;
- update the dropped-worker transport durability test to await the actual `persistCriticalSwarmNow()` flight, then flush/read the command snapshot, instead of yielding arbitrary microtasks and polling.

No production behavior changes.

Closes #52.

## Validation

- the formerly flaky dropped-transport test passed 5/5 repeated focused runs;
- focused bridge/continuation/fault-gate tests passed;
- TypeScript and diff-check passed;
- composed integration verification: 74/74 main test files, 2038 passed + 18 skipped; bridge itself passed 191/191; shutdown 2/2 and production build passed.


totec448-spec:
i let codex analyze all of ur prs detailedly, it doesnt want me to add anythhing before the current 2.0.3 is merged. i will for sure come back and look at the ideas/merge them after the release is done
ChatGPT_ArrI0293Y0

PR #59, opened 2026-09-01
2.0.3: A small step towards the infinite loop

totec448-spec:
Release 2.0.3, built on the squashed 2.0.3 line plus the PR #28 merge already on main.

**A small step towards the infinite loop.**

- **The Loop.** Goal with the exit removed: every finished turn gets its next message until you switch it off.
- **Improved stability and reliability through auto reloads.** Silence, page errors and missing tabs reload or reopen a chat on their own evidence; a turn still running is never closed early.
- **macOS computer use.** The optional Desktop connector ships on macOS, off by default.
- **Auto update.** Windows and AppImage installs stage the next release and apply it on quit.
- **Small things.** UI polish, blocked chats, worker recovery, session timeline fixes.

Release plumbing: version 2.0.3 everywhere the publish preflight checks, changelog head with the full list folded, `docs/release-notes/v2.0.3.md`. Agent handoff notes and a local experiment's proof files are untracked; `npm run verify:privacy` passes on the whole line.


hhh2210:
Mac/Apple Silicon validation from my side on current #59 HEAD (`94d08a2`):

- `npm run build` passes.
- privacy gate and TypeScript typecheck pass.
- full verify reaches 2017 passed / 94 skipped / 1 failed.
- the only failure is the same macOS environment-sensitive `bundled ripgrep` contract I had already noted while validating #28: default login zsh startup rewrites PATH, so `command -v rg` sees `/opt/homebrew/bin/rg` instead of the staged `resources/rg/rg`. Bare `rg` execution is still explicitly bound by `bindBundledRipgrep`, so I am not seeing a Loop/continuation/recovery regression here.

I also hit npm 11 refusing Electron install scripts on the first clean install; after the explicit Electron install the suite reaches the single rg failure above.

Happy to send a very small macOS follow-up if you want the PATH/test contract made consistent before 2.0.3; otherwise I would treat this as known host-environment friction rather than a blocker for the new 2.0.3 features.

PR #68, opened 2026-09-04
spawn: per-worker model and reasoning_effort

JeshuaCastro:
Follow-up to #N.

Root cause: agents spawn plans workers as (label, task) only, so a prime can neither hold back its expensive model nor declare a reasoning level. High-reasoning intent therefore lands on Pro SKUs by default.

Smallest change (implementation + surface + tests + docs, together):
- `src/shared/session.ts` — canonical reasoning vocabulary + `AgentInfo.reasoningEffort` / `model`.
- `src/main/agents.ts` — strict pre-mutation validation of both fields (bad input fails the spawn with zero workers), stored on the worker record, part of the repeat-spawn match, snapshotted and restored.
- `src/main/bridge.ts` — both fields ride the worker's fresh-chat open URL (`?model=` / `?reasoning_effort=`) on the OS-open and extension-placement paths, and the durable command spec.
- `extension/background.js` — placement path appends both, re-validated locally.
- `src/main/mcp/tools-core.ts` — schema, spawn receipts and status show both; `docs/tool-surface.md` documents them.

Validation: `tsc` clean; full vitest gate green except the ripgrep-PATH env test, which also fails on clean v2.0.5 on this machine (Homebrew rg shadows the bundled one — unrelated); packaged arm64 build installed and smoke-checked (bridge listening, no errors). New tests: high-only worker pins no model (URL carries no `model=` param), model+reasoning together, neither pinned, `"banana"` rejected with zero workers, restore round-trip + repair, repeat-folding respects both fields.

One open question: whether chatgpt.com applies `?model=` / `?reasoning_effort=` on fresh chats. Either way the app degrades to defaults, and the opened URL carries both for inspection.

### DIFF
--- a/docs/tool-surface.md
+++ b/docs/tool-surface.md
@@ -106,7 +106,11 @@ Compact & Resume is app/browser orchestration. There is no model-visible `save_h
 Available while multi-agent mode is enabled. It has exactly four actions:
 
 - `spawn` creates worker chats from one shared context plus per-worker tasks. Used once per run:
-  a run that needs a worker again reuses one it already has.
+  a run that needs a worker again reuses one it already has. Each worker takes an optional
+  `model` slug: the worker's chat opens with `?model=<slug>` in its fresh-chat URL, so a prime
+  on a limited model can spawn workers on a cheaper one. Omitted means the account default;
+  a slug ChatGPT does not recognise opens with the default too. The model is fixed for the
+  life of that conversation, including across sleep/wake reuse.
 - `message` sends one message or an all-or-nothing batch. Messaging a sleeping worker is what
   wakes it, in the chat it already has.
 - `status` reports the run and workers, including who is asleep and how many worker slots are free.
--- a/docs/tool-surface.md
+++ b/docs/tool-surface.md
@@ -110,7 +110,10 @@ Available while multi-agent mode is enabled. It has exactly four actions:
   `model` slug: the worker's chat opens with `?model=<slug>` in its fresh-chat URL, so a prime
   on a limited model can spawn workers on a cheaper one. Omitted means the account default;
   a slug ChatGPT does not recognise opens with the default too. The model is fixed for the
-  life of that conversation, including across sleep/wake reuse.
+  life of that conversation, including across sleep/wake reuse. Each worker also takes an
+  optional `reasoning_effort`: none, minimal, low, medium, high, xhigh, max or ultra, forwarded
+  on the open URL independently of `model` — a level never selects or changes the model, and
+  omitting both inherits the normal worker defaults.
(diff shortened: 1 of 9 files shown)


Maximapple:
Heads-up on where this stands after 2.0.6, since a fair amount of it landed by another route.

Much of what this PR describes is on `main` now: the per-worker `model` / `reasoning_effort`
fields on the worker record, strict pre-mutation validation that fails the spawn with zero
workers, both values on the fresh-chat open URL, the snapshot/restore round-trip, repeat-spawn
matching on both fields, receipts and `status`, and the `docs/tool-surface.md` entry. There is
also a new global **"Default sub-agent model"** setting that these override.

What did **not** land is the surface: the `agents` input schema still declared `label` and `task`
only, under `.strict()`, so nothing could actually send either field — and the `task` description
told callers they were fixed in app settings. I have put up #87 for that one gap.

So this PR is not redundant, but it is no longer the same diff. If #87 lands, what is left here is
worth diffing against `main` rather than rebasing wholesale — my guess is most of `agents.ts`,
`bridge.ts` and `session.ts` is already covered and the remainder is small.

On your open question — whether chatgpt.com honours `?model=` / `?reasoning_effort=` on a fresh
chat — `main` did not settle it either. The URL carries both and the app degrades to defaults, and
the app-settings path describes itself as *"Confirmed in ChatGPT before a worker starts"*, which
suggests confirmation happens in the page rather than being assumed from the URL.

Credit where it is due: the design in this PR is the one that shipped.

totec448-spec:

This implementation is superseded by the existing per-worker broker fields, validation, durable transport and receipts plus the schema correction from #87 incorporated through #91. The requested model/reasoning distinction is now exposed on main.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/agents.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/agents.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

totec448-spec:
@JeshuaCastro The per-worker model and reasoning design you contributed here deserved explicit attribution when it was incorporated. Calling this superseded did not adequately recognize that. I am sorry. [CONTRIBUTORS.md](https://github.com/totec448-spec/chat-on-steroids/blob/main/CONTRIBUTORS.md) now credits your design and distinguishes the later schema correction, and [86d7014](https://github.com/totec448-spec/chat-on-steroids/commit/86d7014a2bc18b6dbdf61972ec275e2fc1eb3b40) adds retroactive co-author credit that GitHub links to your account.

Issue #30, opened 2026-08-29
Worker bootstrap fails when New Chat restores an autosaved composer draft


Maximapple:
## Problem

A queued worker bootstrap can fail to start when ChatGPT's **New Chat** composer already contains an autosaved, unsent draft.

This is reproducible when a user types something into a fresh ChatGPT composer, does not send it, leaves the page, and later a Chat On Steroids worker is spawned. ChatGPT restores the saved draft into the fresh-chat composer. The worker tab opens, but the worker bootstrap is never sent.

Observed in the current official build during normal use. I have not yet re-run the exact browser reproduction against my latest local build, but source inspection shows the same guard is still present there.

## Why this happens

Fresh worker/resume commands intentionally open the regular New Chat route:

```ts
const base = conversationId ? `https://chatgpt.com/c/${encodeURIComponent(conversationId)}` : 'https://chatgpt.com/';
return `${base}?${marker}#${marker}`;
```

That route can restore ChatGPT's persisted New Chat draft.

The DOM helper then deliberately refuses to insert a bootstrap whenever the composer is non-empty:

```js
/** Types into the composer. Refuses if the user already has a draft there. */
function insertPrompt(value) {
  return safe(() => {
    const box = composer();
    if (!box) return false;
    if ((box.textContent || '').trim() !== '') return false;
    // ...
  }, false);
}
```

`runCommand()` treats that `false` as a bootstrap failure:

```js
if (!CLF_DOM.insertPrompt(boot.text)) {
  return void (await fail('ChatGPT refused the inserted text'));
}
```

So the current safety rule correctly avoids overwriting a user's draft, but it also means an unrelated saved New Chat draft can prevent workers from starting at all.

## Steps to reproduce

1. Open `https://chatgpt.com/` in a normal New Chat.
2. Type some text into the composer but **do not send it**.
3. Leave/navigate away and confirm ChatGPT restores that draft when returning to New Chat.
4. Start a Chat On Steroids workflow that spawns a worker/sub-agent.
5. The worker opens a fresh ChatGPT tab with the `clf` bootstrap marker.
6. The saved draft is restored into the composer.
7. The bootstrap is not inserted/sent, so the worker does not start.

## Expected behavior

A pre-existing New Chat draft should not silently make worker spawning look like it did nothing.

The default behavior should remain conservative: do **not** silently delete, overwrite, submit, or merge user-authored draft text.

[...]

PR #78, opened 2026-09-05
Add a browser tool: ChatGPT can drive a real web page

Maximapple:
**ChatGPT can now drive a real web browser** — your Chrome, your session, your logins.

---

## 🌐 The `browser` tool

Ask it to fill in a form, work through a checkout, drive an admin panel, or pull a number out of a
dashboard that sits behind a login — and watch it happen in the browser you already have open. No
headless copy. No scraping. No second profile that knows nothing about you.

```
navigate   observe   click_ref   move_ref   set_value   type   scroll   drag
back   forward   reload   status   detach
```

**`observe` hands back the page as things you can name.** Not a DOM dump, not coordinates — a list
of controls with roles, values and readable names, plus a screenshot:

```
g1_e0  combobox "Please select an option Option 1 Option 2" value="1" at 864,100
g3_e0  generic  "name: user1 View profile" at 469,214
g6_e1  link     "Logout" at 431,230
```

Then `click_ref g6_e1`. Refs are re-resolved from the live document at action time, so an element
that moved is still hit and one that vanished is refused — instead of a stale coordinate landing on
whatever slid into that spot, which is the worst outcome available.

**Every action is real input.** `isTrusted: true`, indistinguishable from a person — so pages that
reject synthetic events just work. Trusted clicks, keys, drags, wheel, hover.

**Hover actually works**, which is rarer than it sounds: `move_ref` points at a control without
pressing it, so menus and captions that only exist under the pointer are reachable at all. `observe`
finds them by reading the page's *own* `:hover` rules — the page telling you what reacts to a
pointer.

**Dropdowns work too.** A native `<select>` is painted by the browser, outside the page, where no
synthetic click can ever land — `set_value` picks the option properly and fires `input`+`change`.
Ask for an option that doesn't exist and it says so, and lists the ones that do.

### You can see everything it does

A deliberate design goal, not a nicety:

- the driven tab is collected into a **visibly labelled tab group**
- **Chrome's own debugger banner** stays up — the *browser* tells you, not just the app
- a **pointer overlay** is drawn on the page, so every click and hover has a visible cursor

Nothing the model does to your browser is invisible to you. And `detach` gives the tab back.

### And it can't wander off

ChatGPT's own tabs are refused at attach, at navigate, **and after a click**, so a link can't walk
around the fence. `chrome://`, `file://` and lookalike hosts are judged by parsed identity rather
than string prefix — `https://user@chatgpt.com/` and `https://chatgpt.com.example/` used to be
judged wrongly in opposite directions. Anything that isn't the ordinary web is refused by *not being
on the allowlist*, rather than by someone having remembered to list it.

---

## ⚠️ Recreate the connector before testing

**ChatGPT caches the action list it reviewed**, and the app deliberately doesn't pretend to
hot-patch a cached schema. This adds a whole tool, so an existing connector shows the **old** surface
— the `browser` tool simply won't be there, and it'll look like the PR does nothing.

**Refresh or recreate the custom app in ChatGPT, then start a new conversation.**

---

## 🍎 macOS desktop parity

The macOS connector is yours — it shipped in 2.0.3. This carries over the hardening that had built
up on the Windows side and never crossed:

- **Three refusals named tools that don't exist.** `UNKNOWN_UI_REF`, `STALE_REF` and `STALE_FRAME`
  all said "call `get_window_state` or `find_ui`" — both internal helper ops, neither reachable from
  a tool. The one instruction each refusal existed to give named nothing you could act on. They now
  say `observe`.
- **`BAD_KEY` said only "unknown key"**, where Windows has listed its valid keys all along. Both list
  them now, each from its own table — the platforms genuinely differ.
- **UI-snapshot history was 16** — the number Windows had already *measured* as too few, after
  thirteen workers sharing a helper evicted each other's snapshots into forty stale refusals in one
  run. Both are 96, pinned together by a test.

---

## ✨ And a couple of dozen fixes

Most share one shape — *the call succeeded and nothing happened* — which is the single failure a
tool user can't debug:

- **A click that reaches nothing now says so.** Chrome's password-breach dialog sits *above* the
  page, painted by the browser, silently eating input. `elementFromPoint` can't see browser UI, so
  the click was dispatched, reported trusted, and landed nowhere. The dialog can't be detected — its
  *effect* can, so the answer now reports the page didn't move and names the address it never
  reached.
- **Chrome's error page was mistaken for the site.** Three APIs answer "what is this tab showing"
  and two of them lie about a failed load: `chrome.tabs` and `Page.getNavigationHistory` both name
  the address that was *requested*. Only the frame tree admits `chrome-error://chromewebdata/`.
  Every guard was asking a lying surface.
- **A stalled Compact & Resume took browser recovery down with it** — for six hours, silently. Two
  separate routes into that state, both closed.
- **"Waiting for 2 tool calls" outlived the app running them.** A restart left the panel asserting a
  dead process's state for ten minutes.
- **Refusals that ask for something now name it.** `WORKSPACE_REQUIRED` lists the approved roots;
  the input fences point at the `browser` tool when the target is a browser window — a sentence that
  had lived only in a source comment. No fence was weakened to do it.

## Testing

Four QA rounds on macOS and through ChatGPT's own connector, on real pages and a running app. Every
fix carries a regression test confirmed to fail without it, and three harnesses drive **real Chrome**
rather than mocks: `verify:browser` (74 checks), `verify:compact-chain`, and
`verify:compaction-release` — the last confirmed in both directions, with and without its fix.

`tsc` clean, 2411 tests, privacy gate green, six-platform build green.
`docs/qa-round-2026-09-05-disposition.md` has every finding with its cause and evidence.

## One note for the reviewer

Upstream's `qualifiedFrame(…, generationOfReply(reply))` and this branch's bounds check landed on
the same line and both are kept — "is this the right frame" and "is the point inside it" are
different questions. On the QA hardware the frame was current, so the bounds check was the only
thing between the caller and a coordinate that can't exist. Upstream's own test for that code passes
against the merged form; the bounds half was driven on hardware, the generation half read (the macOS
helper is an in-process addon, so a stale generation can't be induced from outside).


Maximapple:
Rebuilt on 2.0.6. Congratulations on the release — and this branch is the one that had to move,
not you.

Twenty-eight conflicting hunks across seventeen files. Most were textual, but five were a real
choice about which behaviour survives, and rather than leave you to reverse-engineer those from
the diff, here is each one and why it went the way it did. **In four of the five, 2.0.6 won.**

### `ButtonFlags` keeps both fixes

2.0.6 shipped #79's swapped-button fix; this branch had added the side buttons, which are one
event pair distinguished by `mouseData` rather than by flags of their own. Merged so that both
hold: the swap is read per call and applied to the primary/secondary pair only, while `back` and
`forward` keep their own word untouched. `SM_SWAPBUTTON` is a checkbox in Mouse properties and can
change while the app runs, so it is still read per call rather than cached.

### The prime workspace takes yours

This branch had deleted the `agent:prime` mirror so a later run could not inherit an earlier one's
folder without either conversation ever naming it. 2.0.6 reaches the same place more thoroughly:
every staged key is now run-scoped (`agent:<runId>:<agentId>`), which closes that leak for primes
and workers alike instead of only for the prime. Yours is the better mechanism, so it wins
outright and your tests replace this branch's.

### The worker-draft preference is dropped

This branch carried the opt-in setting proposed in #30 — a popup toggle, a storage key, its own
test. 2.0.6 settled the same question as an unconditional replace, and reintroducing an opt-in
that defaults to *preserving* would put back exactly the failure `3bdc3e2` removed: a worker that
refuses to start because ChatGPT restored a New Chat draft. That is the worse of the two failures,
so the setting is gone from this branch entirely and #30 stays yours to decide on its own terms,
not as a side effect of a browser-tool PR.

### The privacy gate takes yours, and this branch's relaxation is gone

Both sides had solved "a fork must be able to inherit already-public history". This branch did it
by switching the historical identity check off when the checkout was not the maintainer's own
repository. 2.0.6 does it by finding the canonical repository *by URL* under whatever name the
remote carries, and exempting only what is actually published there.

Yours is strictly safer and this one should not have shipped: relaxing by owner turns the check
off in the fork, which is the checkout most likely to leak in the first place. Removed, with the
`open-issues-hardening` assertion rewritten to pin the replacement rule so it cannot come back.
The blocked-value additions from this branch are kept — they were each added after a real value
reached a report.

### `listAllSessions()` is not reintroduced

It wraps the capped `readAllSummaries()`, which this branch deleted after an audit found it being
read as authoritative — the cap silently turns "session 5,001" into "does not exist" for callers
whose whole contract is not to do that. Nothing calls it in the merged tree, so it stays out.
`listUsageSessions()` is untouched and still backs the new usage view.

### One harness change worth flagging

`test/desktop-input-maintenance.test.ts` evaluates `background.js` as a classic script, and this
branch made the browser-driver import static. It is static on purpose — when it was dynamic, every
`browser_*` message failed in a real browser while the tests stayed green, because they never
execute one. So the harness now strips the import and supplies the binding, exactly as
`test/extension.test.ts` already did, and throws if that import ever stops looking the way the
rewrite expects. The alternative — making the worker import dynamically to suit a test — would
reintroduce the silence.

### Verification

`npm run typecheck` clean. 3108 tests passing.

Two failures remain, and both reproduce identically on a **pristine `main` checkout in the same
environment**, which is how they were classified rather than assumed: PowerShell execution policy
blocking `.ps1` (`test/mcp.test.ts`), and a locale that formats decimals with a comma, so `0,48`
does not contain `0.48` (`test/renderer-usage.test.ts`). Neither is caused by this branch — and
the merge incidentally fixed a third baseline failure, in the batch-scoped exec recovery hints.

The browser tool itself is unchanged by all of this and still has no counterpart on `main`, so
nothing in the feature needed re-litigating — only the seams around it.

totec448-spec:
Leaving this unmerged pending a focused split. The diff against current main spans 158 files and over 32,000 added lines, including browser control, native/exec/session changes and extensive historical QA records. That is not a reviewable browser-only change and must not publish unrelated implementation or local QA history. Please extract the browser capability with its exact tab ownership, live capability enforcement and bounded output tests, and submit unrelated native/session fixes independently. This is a scope/acceptance decision, not a claim that every added line is defective.

Maximapple:
Split as asked. Closing this in favour of the pieces.

**The browser capability, on its own: #142**

| | this PR | #142 |
|---|---|---|
| files | 156 | 20 |
| insertions | 31,773 | 3,746 |
| `docs/` | 45 files, 13,804 lines | none |
| native / session / scripts | yes | none |

It carries the three things you named, and the main structural change is that the tool now lives in its own module rather than threaded through `tools-desktop.ts`, so tab ownership, the live permission gate and the bounded output are reviewable together. Discovery cost is stated rather than absorbed: Desktop's `tools/list` goes 7.9k → 12.9k bytes, with the surface ceiling and a per-tool ceiling both asserted.

**Unrelated fixes, separately:**

- #139 — the three callers that promise to see every retained session were reading the bounded list. One of them is the request-correlation reconciliation added in #134, and `readAllSummaries()` says in its own doc comment not to use it for identity.
- #141 — `destinationLost` is sent by `content.js` and acted on by `bridge.ts`, and dropped by the relay in between, so the "re-offer the brief at once" path cannot run.

More to follow from the same branch (the native desktop work, the session-store audit). Each will stand on its own.

The QA history stays out of all of them — you were right that it does not belong in a published diff.

Thanks for the specific list; it made the cut obvious rather than a guess.

PR #80, opened 2026-09-05
Give the macOS app a coherent ad-hoc seal so it will launch

### LINKED ISSUE #66: macOS arm64 DMG 2.0.5: Finder reports app is damaged (broken ad-hoc seal) on macOS 27, MacBook Air M4
## App version
2.0.5

## Operating system and version
macOS 27.0 (Build 26A5425a), MacBook Air M4, installed from Chat-On-Steroids-macOS-arm64.dmg into /Applications

## Architecture
ARM64

## Was the Chrome extension connected?
No

## ChatGPT workspace / custom MCP access
Not relevant to this bug

## What happened
After installing the arm64 DMG into /Applications, double-clicking Chat On Steroids in Finder shows the macOS damaged dialog (move to Trash). The app never starts, so no in-app logs exist.

## Steps to reproduce
1. Download Chat-On-Steroids-macOS-arm64.dmg (2.0.5).
2. Open the DMG, copy Chat On Steroids.app to /Applications.
3. Double-click it in Finder.

Expected: app launches. Actual: macOS reports it is damaged and offers to move it to Trash.

## Anything else
Local diagnostics (paths sanitized):

- No crash reports for the app in ~/Library/Logs/DiagnosticReports or /Library/Logs/DiagnosticReports.
- Direct execution dies immediately:
  /Applications/Chat On Steroids.app/Contents/MacOS/Chat On Steroids -> EXIT:137 (SIGKILL), no output.
- codesign --verify fails identically on the installed app AND on the app inside the mounted DMG, so it is broken in the release artifact, not by copying:
  codesign --verify --deep --strict --verbose=2 "/Applications/Chat On Steroids.app"
  -> Chat On Steroids.app: code has no resources but signature indicates they must be present
- codesign -dvv: Format=app bundle with Mach-O thin (arm64), flags=0x20002(adhoc,linker-signed), Signature=adhoc, TeamIdentifier=not set, Sealed Resources=none.
- Same verify failure on the nested Electron Framework.framework.
- Quarantine flag present from download: com.apple.quarantine: 0181;...;Chrome;....
- spctl --status on this machine: assessments disabled, yet Finder still blocks it (structurally invalid seal, not policy).

Repo context: electron-builder.yml sets mac identity: null + notarize: false (intentionally unsigned per comment), and scripts/smoke-macos-bundle.mjs deliberately avoids codesign --verify --deep, asserting only no trust-bearing signature. CI launches the unpacked binary directly (no quarantine / LaunchServices path), so this Finder/Gatekeeper failure on a newer macOS is not covered.

Possible directions (maintainer call): real Developer-ID signing + notarization; or re-seal ad-hoc after build (codesign --force --deep --sign -) plus verify; or strip signatures + document quarantine removal. Happy to test a candidate DMG/ZIP on this machine.

Maximapple:
Fixes #66 — the arm64 DMG opens to "the application is damaged" and never launches.

## The diagnosis

Gatekeeper assessment was already disabled on the reporting machine, so this was structural rather
than policy:

```
codesign --verify --deep --strict "Chat On Steroids.app"
-> code has no resources but signature indicates they must be present
codesign -dvv -> flags=0x20002(adhoc,linker-signed), Sealed Resources=none
```

`linker-signed` is the whole story. **Every arm64 Mach-O gets an ad-hoc signature from the linker
whether anyone asks or not** — it isn't optional on Apple Silicon. `identity: null` skips *bundle*
signing, so the app shipped as a bundle full of signed executables with no
`_CodeSignature/CodeResources`. macOS reads the signature on the executable, looks for the resource
seal it implies, finds none, and calls the bundle damaged.

## The fix

An `afterPack` hook re-seals the bundle ad-hoc and then verifies it.

**This does not weaken the unsigned policy.** An ad-hoc seal carries no `Authority` and no
`TeamIdentifier`; `codesign -dvv` still reports `Signature=adhoc`, and Gatekeeper still won't vouch
for it. It adds only the envelope the executables already claim exists — the bundle stops
contradicting itself. The hook fails the build if it ever produces something trust-bearing, so it
can't quietly turn into a signing change.

## Why it shipped twice

This is the part worth a look, because it wasn't really a packaging bug — it was an assertion:

```js
if (hasCodeResources) {
  throw new Error(`${file} unexpectedly has a bundle CodeResources signature envelope`);
}
```

The bundle audit **threw when the seal existed**, so the broken artifact was the compliant one and
nothing could ever catch it. That's now inverted — the seal is required — and the test grew a case
for the exact shape that shipped: ad-hoc executables, no resource envelope.

Trust and coherence are different axes. Only the first is the release policy, and it's still
enforced.

## What I could not check

I'm on Windows; `codesign` doesn't exist here. The hook is written from the report's diagnosis,
which is unusually complete, and the `--verify --deep --strict` step is deliberately *inside* the
build so a bad seal fails there rather than reaching a user.

@lavalava45 offered to test a candidate DMG on the affected machine — that's the check this wants
before it goes into a release, and it's the one thing I can't do from here.

`tsc` clean, suite green. (`exec-hints.test.ts` fails on my machine from PowerShell's execution
policy blocking `.ps1` files; it fails identically on a clean checkout of `main`.)

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -34,6 +34,9 @@ extraResources:
 
 npmRebuild: false
 
+# macOS only; a no-op on every other platform. See the hook and the mac.identity note below.
+afterPack: scripts/afterpack-macos-adhoc-seal.mjs
+
 asarUnpack:
   - '**/node_modules/node-pty/**'
   - '**/node_modules/sharp/**'
@@ -90,6 +93,12 @@ mac:
   # v2.0.2 intentionally ships unsigned/unnotarized macOS artifacts. Make that policy
   # deterministic instead of depending on whatever identities or Apple env vars happen to be
   # present on a hosted runner. Remove these two lines when real signing credentials are wired.
+  #
+  # Unsigned still has to mean *coherent*. Every arm64 Mach-O is ad-hoc signed by the linker
+  # whether anyone asks or not, so skipping bundle signing shipped an app whose executables
+  # claimed a resource seal the bundle did not have; macOS calls that damaged and refuses to
+  # launch it (issue #66). afterPack re-seals ad-hoc — no Authority, no TeamIdentifier, still
+  # nothing Gatekeeper will vouch for — and verifies the result.
   identity: null
   notarize: false
   artifactName: Chat-On-Steroids-macOS-${arch}.${ext}
--- a/scripts/afterpack-macos-adhoc-seal.mjs
+++ b/scripts/afterpack-macos-adhoc-seal.mjs
@@ -1,0 +1,76 @@
+/**
+ * Gives the packaged macOS app a valid ad-hoc seal, and refuses to ship one that is not valid.
+ *
+ * ## The bug this exists for
+ *
+ * Issue #66: the 2.0.5 arm64 DMG launched to "the application is damaged" on macOS 27, with
+ * Gatekeeper assessment already disabled — so this was a structural failure, not a policy one:
+ *
+ *     codesign --verify --deep --strict "Chat On Steroids.app"
+ *     -> code has no resources but signature indicates they must be present
+ *     codesign -dvv -> flags=0x20002(adhoc,linker-signed), Sealed Resources=none
+ *
+ * `linker-signed` is the giveaway. Every arm64 Mach-O gets an ad-hoc signature from the linker
+ * whether anyone asks or not — it is not optional on Apple Silicon. `identity: null` tells
+ * electron-builder to skip bundle signing, so the app shipped as a *bundle* carrying signed
+ * Mach-Os and no `_CodeSignature/CodeResources`. macOS reads the signature on the executable,
+ * looks for the resource seal it implies, finds none, and calls the bundle damaged. There was
+ * never an x64-only version of this problem, which is why it appeared with Apple Silicon.
+ *
+ * ## Why ad-hoc rather than Developer ID
+ *
+ * The release policy is deliberately unsigned and unnotarized, and this does not change that.
+ * An ad-hoc seal carries no Authority and no TeamIdentifier — `codesign -dvv` still reports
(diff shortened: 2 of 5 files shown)


Maximapple:
Superseded by 2.0.6 — closing.

The release commit carries this work already, so there is nothing here left to review:

- `scripts/macos-audit-utils.mjs` on `main` is byte-identical to this branch's, including the
  inverted seal check that was the actual root cause: the audit demanded the broken state, which
  is how two releases shipped a bundle macOS reads as damaged.
- `scripts/afterpack-macos-adhoc-seal.mjs` is this branch's script with three changes, and one of
  them is a real bug fix that belongs to whoever made it.

That fix is worth stating plainly, because this branch was wrong. `run()` here used
`execFileSync`, whose return value is **stdout only**, and `codesign --display --verbose=4` writes
its details to **stderr**. So `shown` would have been empty, `/^Signature=adhoc\s*$/m` would not
match, and every macOS build would have failed at the afterPack step with a message quoting no
output at all. `main` switched to `spawnSync`, kept the whole result, and routed it through
`assertNoTrustBearingMacCodeSignature`, which reads `stdout + stderr`, distinguishes a non-zero
exit from an unsigned object, and also rejects a `TeamIdentifier`. That is strictly better than
what this PR proposed, and it also removes the duplicated policy that lived in both files.

`test/macos-adhoc-seal.test.ts` is on `main` too, and this branch never had it.

One thing this PR could not do and still cannot: **nothing here has ever run `codesign`.** CI
verifies the scripts, not the seal, so the closing evidence for #66 is still a real
`.dmg` from a macOS build that launches on a machine that did not produce it. `main` now also runs
`codesign --verify --deep --strict` against the copied payload in the bundle smoke check, which
covers the "DMG construction dropped a sealed resource" case that a source-side verify cannot see.

@lavalava45 — if the offer to test a candidate build still stands, a 2.0.6 arm64 `.dmg` opened on
a machine that did not build it is the check that would actually close #66.

PR #86, opened 2026-09-06
Create the Compact & Resume successor inside its source Project

### LINKED ISSUE #84: Compact & Resume opens replacement chat outside the source ChatGPT Project
### App version

2.0.5 / current main

Windows 11 Pro 24H2 (build 26100)

### Architecture

x64 (Intel/AMD)

Not relevant to this bug

### What happened

When Compact & Resume is started from a conversation inside a ChatGPT Project (`/g/<project>/c/<conversation>`), the replacement conversation is created at the site root instead of inside the same Project.

Current main builds the fresh resume URL as `https://chatgpt.com/?clf=...` in both the extension placement path and the OS fallback. As a result, chat B is created outside the Project even though chat A belonged to one.

Expected: a replacement created by Compact & Resume should keep the source conversation's Project membership. Non-Project chats should continue to open at the root.

### Steps to reproduce

1. Open a recorded ChatGPT conversation inside a Project (`https://chatgpt.com/g/<project>/c/<conversation>`).
2. Trigger **Compact & Resume** and let the handoff complete.
3. Let the extension open the replacement chat.
4. Observe that the replacement starts from `https://chatgpt.com/?clf=...`, so the resulting conversation is created outside `/g/<project>/...`.

The OS fallback has the same problem because `commandUrl()` also returns the root New Chat URL.

### Anything else

```text
This appears separate from #25 / #38. Those fixed recognition and caller attribution for existing Project conversation routes (`/g/<project>/c/<id>`); this issue is about preserving Project creation affinity when Compact & Resume creates the successor chat.

Implementation suggestion:
- Capture a validated Project route/slug from chat A when the continuation is opened.
- Persist it with the continuation so restart/recovery does not lose it.
- Use it for both the extension placement URL and the OS fallback URL (for example `/g/<project>/project?clf=...`).
- Keep old continuation records and non-Project chats backward-compatible.

A browser-only fix in `placeSuccessorChat()` would cover the happy path but would still lose the Project on OS fallback/recovery.

Regression tests should cover extension placement, OS fallback/restart recovery, unchanged non-Project behavior, and invalid Project route input.
```

Maximapple:
Fixes #84.

A ChatGPT Project is expressed only in the address. `/g/<project>/c/<id>` is a chat filed under
one, and a conversation created from anywhere else is not in it — no parameter, header or later
call moves it there. So the successor's URL *is* the mechanism, and all three places that build
one opened at the site root:

| | |
|---|---|
| `extension/background.js` `placeSuccessorChat()` | window-bound placement — the path Compact & Resume actually takes |
| the same function's `background: true` branch | a worker chat in the minimized window |
| `src/main/bridge.ts` `commandUrl()` | the app's OS fallback, and restart recovery |

### How the Project gets from chat A to chat B

Only the browser can see it, and only the app survives a restart, so it travels between them:
the page reports it on the `/activity` poll it already makes, the app writes it onto the
continuation, and all three builders read it from there. The extension prefers the value the app
names — that is the one that outlives a tab — and falls back to reading the tab it is placing
beside. The OS fallback reads the continuation, because by the time it runs there may be no tab
at all.

### The slug is not what gets carried

This is the part worth reviewing closely. A Project chat's path has the Project's **current
display name appended to its id**:

```
chat in a Project   /g/g-p-<32 hex>-homelab-development/c/<conversation>
the Project itself  /g/g-p-<32 hex>/project
```

Copying the slug out of chat A verbatim would build an address containing a name that a rename
invalidates. Only the `g-p-<hex>` head — what `/g/<id>/project` is addressed by — is ever stored
or emitted, and normalization happens at the boundary so a display-name form cannot reach a
durable record even if something offers one.

### Nothing is guessed

Anything that is not exactly `g-p-` plus 32 hex digits opens at the root, exactly as before: a
record written by an older build, a corrupt or tampered one, a custom GPT's `/g/` slug (custom
GPTs are served from `/g/` too and are not Projects), a display-name form, a path-escape attempt.

Degrading to the previous behaviour is deliberate. A Project this app does not recognise costs
the user what they have today; an address assembled from a value it does not understand could
cost them something worse.

### Verification

25 tests in `test/project-successor.test.ts`, covering the five cases the issue asks for:
extension placement (both windows), OS fallback, restart recovery, unchanged non-Project
behaviour, and rejected route input — plus old-record and corrupt-record migration.

The placement tests drive the **real service worker** in a VM and read the tab it creates, rather
than asserting on source text. Checked the way that matters: they fail against the previous
`background.js` and pass against this one.

`npm run typecheck` clean. Full suite: 2946 passing. Two failures remain and both reproduce
identically on a pristine `main` checkout in the same environment — PowerShell execution policy
blocking `.ps1`, and a locale that formats decimals with a comma.

### One thing I could not verify from here

The route shapes are taken from a real account rather than invented, and both are in the tests as
constants. What is **not** machine-verified is the last step: that opening `/g/<id>/project?clf=…`
and sending actually files the new conversation under that Project. It is the address ChatGPT
serves the Project page from, and the bootstrap is path-agnostic — the marker is read from the
query and fragment, and `conversationFromPath()` correctly reports no conversation there, exactly
as at the root — so there is no reason it should behave differently. But it wants one real
Compact & Resume from inside a Project to confirm, and I would rather say so than imply otherwise.

If it turns out ChatGPT needs a different entry point, only `successorChatBase()` and
`commandUrl()` change; the capture, persistence and fallback behaviour are independent of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ The app and the `extension/` companion are versioned together. **Reload the
 extension after updating the app**. If their bridge protocols are incompatible,
 the app refuses the extension and asks you to reload the matching copy.
 
+## Unreleased
+
+### Fixed
+- **Compact & Resume no longer drops the replacement chat outside its Project.** Starting a
+  handoff from a conversation inside a ChatGPT Project created the successor at the site root,
+  so the work continued outside the Project it belonged to. A Project is expressed only in the
+  address — `/g/<project>/c/<id>` — and a chat created from anywhere else is not in it, so the
+  successor's URL was the entire mechanism, and all three places that build one were hardcoded
+  to the root: the extension's window-bound placement, its background-window placement and the
+  app's own OS fallback. The Project is now read from the polling page's own address, written
+  onto the continuation so a restart or a closed tab cannot lose it, and used by all three.
+
+  Two things this deliberately does not do. It does not carry the slug it sees: a Project chat's
+  path has the Project's display name appended to its id, so only the `g-p-<id>` head is stored
+  and emitted, and renaming a Project cannot invalidate a stored address. And it never guesses —
+  a value that is not exactly that shape, including one from a record written before this
+  existed, opens at the root exactly as before rather than at an address assembled from it.
+
 ## [2.0.6] — 2026-09-06
 
 **I am exhausted.**
--- a/extension/background.js
+++ b/extension/background.js
@@ -2236,6 +2236,27 @@ function conversationFromUrl(value) {
   }
 }
 
+/**
+ * The ChatGPT Project a URL belongs to, or null.
+ *
+ * Mirrors chatgpt-dom.js::projectFromPath and src/main/session/continuation.ts's
+ * normalizeProjectId: only `g-p-` plus 32 hex digits counts. A Project chat's path appends the
+ * Project's display name to that id, so the name is stripped here rather than carried into an
+ * address that a rename would invalidate. Custom GPTs are also served from `/g/`, and their
+ * slugs do not have this shape, so they are correctly not Projects.
+ */
+function projectFromUrl(value) {
+  try {
+    const url = new URL(String(value || ''));
+    if (url.protocol !== 'https:' || (url.hostname !== 'chatgpt.com' && url.hostname !== 'chat.openai.com')) return null;
+    if (url.pathname.length > 512) return null;
+    const match = /^\/g\/(g-p-[0-9a-f]{32})(?:-[^/]*)?\//i.exec(url.pathname);
+    return match ? match[1].toLowerCase() : null;
+  } catch {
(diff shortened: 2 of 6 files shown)


totec448-spec:

Project-preserving successors are incorporated. The adaptation captures Project identity at the existing source-send admission and persists it with the continuation, replacing poll-derived ownership.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/session/continuation.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/project-successor.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

PR #87, opened 2026-09-06
Expose per-worker model and reasoning on the agents schema

### LINKED ISSUE #67: Prime can't choose worker models or reasoning levels
A prime on a limited model (e.g. Pro with capped messages) spawns workers that inherit whatever the account default is, with no way to pick a cheaper model or a reasoning level per worker. The `workers[]` schema offers label + task only, so "high-reasoning subagent" intent has nowhere to go except a more expensive model — the exact wrong mapping.

Proposal: two optional per-worker fields, forwarded independently.

workers: [{
  label?: string
  task: string
  model?: string            // slug; omitted = account default
  reasoning_effort?: string // none, minimal, low, medium, high, xhigh, max, ultra; omitted = inherit
}]

- `model` selects which model the worker's fresh chat opens with.
- `reasoning_effort` declares how much reasoning, and must never select or change the model.
- Unknown values fail the spawn before anything is created — no silent downgrade to a default, or status will claim one thing while the worker does another.
- Both visible in spawn receipts and status; omitting both keeps today's behavior exactly.

Maximapple:
Fixes #67.

Almost all of this already exists on `main`. What is missing is the only part a caller can see.

### What is already there

`planSpawn` in `src/main/agents.ts` reads `worker.model` and `worker.reasoning_effort` **per
worker**, falling back to the app-settings default when they are absent:

```ts
const model = normalizeModel(index, worker.model === undefined ? getConfig().multiAgent.defaultModel : worker.model);
const reasoningEffort = normalizeReasoningEffort(index, worker.reasoning_effort === undefined ? getConfig().multiAgent.defaultReasoning : worker.reasoning_effort);
```

`normalizeModel` and `normalizeReasoningEffort` throw an `AgentError` naming the worker index
before anything is created — so an unrecognised value fails the whole spawn with zero workers,
which is exactly what the issue asked for and is already covered by tests
(`rejects an unknown reasoning_effort with zero workers created`, and the model equivalent). Both
values ride the worker's fresh-chat URL and appear in spawn receipts and `status`.
`docs/tool-surface.md` documents both fields as available.

### What was wrong

The `workers` object in the `agents` input schema is `.strict()` and declared `label` and `task`
only. So a spawn naming a model was **rejected by the schema** rather than honoured — and the
`task` description told the caller the opposite:

> *"Model and reasoning are predefined by the user in app settings."*

Documentation, implementation and tests all said this feature existed. The surface said it did
not, and the surface is what the model reads.

### The change

Both fields are declared on the schema, and that sentence is removed from `task`.

`reasoning_effort` is an `enum` over `REASONING_EFFORTS` rather than a free string, so a caller
can discover the levels instead of guessing at one and having its spawn fail. That constant is
already described in `shared/session.ts` as the single vocabulary authority the broker, the
bridge restore and the extension all check against — using it here means the surface cannot drift
from what is accepted. `model` stays a bounded free slug, validated where it always was.

Omitting either field is unchanged in every respect: the default the user chose in settings, or
the account default when they chose none. Nothing in the broker moved.

The documented level list gains `pro`, which the broker has always accepted and the docs did not
mention — a worker is a real ChatGPT browser chat, so the browser Power tier belongs in its
vocabulary.

### Verification

`npm run typecheck` clean. 2922 passing. The two failures left are environmental and reproduce
identically on a pristine `main` checkout here: PowerShell execution policy blocking `.ps1`, and
a locale that formats decimals with a comma.

The new test asserts the *surface* rather than the broker, since that gate is the one seam the
broker's own tests cannot see — required fields, `additionalProperties: false`, the enum matching
the shared vocabulary, and that the task description no longer contradicts it. It fails against
the previous schema with `expected [ 'label', 'task' ] to deeply equal [ 'label', 'model', … ]`.

### On #68

@JeshuaCastro's PR #68 proposed this feature end to end and now conflicts with `main`. Much of
what it described — the spec fields, the URL transport, the receipts, the strict pre-mutation
validation — is on `main` already, so if this lands, what remains of #68 is worth checking
against it rather than rebasing wholesale.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ The app and the `extension/` companion are versioned together. **Reload the
 extension after updating the app**. If their bridge protocols are incompatible,
 the app refuses the extension and asks you to reload the matching copy.
 
+## Unreleased
+
+### Fixed
+- **A prime can choose each worker's model and reasoning level again.** The broker has accepted
+  per-worker `model` and `reasoning_effort` for some time — validating both before creating
+  anything, failing the whole spawn on an unrecognised value, carrying them on the worker's
+  fresh-chat URL, and reporting them in spawn receipts and status — and `docs/tool-surface.md`
+  documented them as available. The `agents` input schema never exposed them: its `workers`
+  object is `.strict()` and declared only `label` and `task`, so a spawn naming a model was
+  rejected outright rather than honoured, and the task description told the caller the opposite,
+  that model and reasoning were fixed in app settings. Both fields are now on the schema, with
+  `reasoning_effort` declared as the canonical vocabulary so a caller can discover it instead of
+  guessing. Omitting them is unchanged: the default from app settings, or the account default.
+
 ## [2.0.6] — 2026-09-06
 
 **I am exhausted.**
--- a/docs/tool-surface.md
+++ b/docs/tool-surface.md
@@ -111,9 +111,11 @@ Available while multi-agent mode is enabled. It has exactly four actions:
   on a limited model can spawn workers on a cheaper one. Omitted means the account default;
   a slug ChatGPT does not recognise opens with the default too. The model is fixed for the
   life of that conversation, including across sleep/wake reuse. Each worker also takes an
-  optional `reasoning_effort`: none, minimal, low, medium, high, xhigh, max or ultra, forwarded
-  on the open URL independently of `model` — a level never selects or changes the model, and
-  omitting both inherits the normal worker defaults.
+  optional `reasoning_effort`: pro, none, minimal, low, medium, high, xhigh, max or ultra,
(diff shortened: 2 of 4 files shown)


totec448-spec:

The missing public agents schema fields are incorporated. Workers can independently name model and reasoning_effort through the existing broker validation and receipts.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/mcp/tools-core.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/mcp.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

PR #88, opened 2026-09-06
Add a settings field for your own connector instructions

### LINKED ISSUE #85: Add custom MCP instructions setting
### Problem or workflow

There is no persistent way to add my own instructions to the Core/Desktop MCP context. Editing the packaged app is overwritten by updates.

### Proposed change

Add a settings field for custom MCP instructions and append it to the existing Core/Desktop server instructions. It should persist across app updates.

### Alternatives considered

Patching `app.asar` manually.

Maximapple:
Fixes #85.

A settings field whose text is appended to what the Core and Desktop MCP servers tell ChatGPT
about themselves. Stored in config under `mcp.instructions`, so an app update no longer
overwrites it — patching `app.asar` was previously the only way to add anything at all.

### The two decisions worth reviewing

**It goes last.** Everything above it is what the app can actually promise about its own tools.
A preference placed before that could be read as qualifying guidance the app has to honour, so
the user's text can only ever follow it. There is a test asserting the connector's own
description is unchanged character for character and only the new block is appended.

**It is attributed**, under a heading naming the user as its author:

```
The user's own standing instructions for this connector:
Always run the test suite before saying a change works.
```

The model should be able to tell a standing instruction from this person apart from the
connector's description of itself. Those carry different authority, and running them together
hides that. Empty is the normal case and emits nothing at all — not even the heading.

### Reconnect behaviour

Editing this now counts as a tool-surface change, so it raises the same *"Tools changed. Start a
new ChatGPT conversation…"* notice that adding a tool does. ChatGPT reads connector instructions
once, when it loads the tools; without that the edit would look applied while nothing had
actually changed for the running conversation.

### Failure handling

The config field is repaired rather than rejected, exactly as the Goal prompts directly above it
are — free text a person typed, capped at 4000 characters and trimmed. One over-long field must
not send every root and every permission through conservative recovery. There is a test for that
which also asserts the rest of the config survives intact.

The renderer reads the group defensively (`config.mcp?.instructions ?? ''`). A state push from an
older main process carries no `mcp` at all, and a missing settings group should degrade to "none
set" rather than throw through `apply()` and blank the panel. Found by the existing renderer
tests, whose fixtures do exactly that.

### Verification

`npm run typecheck` clean. 2928 passing. The two failures left are environmental and reproduce
identically on a pristine `main` checkout here: PowerShell execution policy blocking `.ps1`, and
a locale that formats decimals with a comma.

Seven new tests in `test/mcp-user-instructions.test.ts`: empty adds nothing on either surface,
appended last and attributed on both, the app's own text untouched, survives a config reload,
whitespace-only trims to nothing, and an over-long value is capped without damaging the config.

### Open to direction

Placement and wording are yours to move — I put it under its own **Connector instructions**
heading in settings, but it would sit equally well beside the connector cards in the wizard. The
heading text itself is in one place (`userInstructions()` in `instructions.ts`) if you would
rather word it differently, and it is deliberately plain rather than a fence, since this is the
user's own app and their own words, not untrusted input.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ The app and the `extension/` companion are versioned together. **Reload the
 extension after updating the app**. If their bridge protocols are incompatible,
 the app refuses the extension and asks you to reload the matching copy.
 
+## Unreleased
+
+### Added
+- **Your own instructions for the connectors.** A settings field whose text is appended to what
+  the Core and Desktop MCP servers tell ChatGPT about themselves, so a standing preference — run
+  the tests before claiming a change works, prefer this package manager, never force-push — does
+  not have to be repeated in every chat. It is stored in settings, so an app update no longer
+  overwrites it; editing `app.asar` was previously the only way to add anything at all.
+
+  It is appended last and under a heading naming the user as its author. Both are deliberate:
+  everything above it is what the app can actually promise about its own tools, and a preference
+  must not silently redefine one — and the model should be able to tell a standing instruction
+  from the person apart from the connector's description of itself, because those carry different
+  authority. Empty adds nothing, not even the heading. Changing it asks for the same reconnect a
+  tool change does, since ChatGPT reads connector instructions once, when it loads the tools.
+
 ## [2.0.6] — 2026-09-06
 
 **I am exhausted.**
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -238,6 +238,16 @@ const capabilitiesSchema = z
   )
   .transform((caps) => ({ ...DEFAULT_CAPABILITIES, ...caps }) as Capabilities);
 
+/**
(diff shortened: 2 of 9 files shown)


totec448-spec:

Custom user-authored connector instructions are incorporated, including persistence, settings merge handling and attribution in the Core/Desktop instruction text.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/mcp/instructions.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/mcp-user-instructions.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

Issue #90, opened 2026-09-06
Chrome extension repeatedly opens Plugin settings when ChatGPT Business has no Refresh button


TaeyanG4:
## Summary

After installing/enabling the Chat On Steroids Chrome extension, ChatGPT repeatedly opens the settings page for **Plugins → Chat On Steroids Core**.

The Core connector itself is working correctly: file reads, file writes (`apply_patch`) and terminal commands (`exec_command`) all work through the OpenAI Secure MCP Tunnel. The repeated settings navigation begins only after the Chrome extension is installed/enabled.

On this ChatGPT Business workspace, the plugin detail page shows the connector actions and input schemas, but there is **no visible `Refresh` button**.

## Environment

- Chat On Steroids: 2.0.6, running from source with `npm run dev`
- OS: Windows
- Browser: Google Chrome
- ChatGPT workspace: Business
- Core connector: created, published, installed, and confirmed working
- Chrome extension: unpacked extension from this repository

## Steps to reproduce

1. Configure Chat On Steroids Core with an OpenAI Secure MCP Tunnel.
2. Create/publish the `Chat On Steroids Core` custom app in ChatGPT Business.
3. Confirm the connector works by calling `read`, `apply_patch`, and `exec_command`.
4. Install/enable the Chat On Steroids Chrome extension.
5. Keep ChatGPT open.
6. Observe that ChatGPT repeatedly navigates to / opens the Core plugin settings page.

The generated navigation appears to be the plugin-refresh workflow, with a URL shaped like:

```text
https://chatgpt.com/?cos-plugin-refresh=<id>#settings/Plugins/...
```

## Actual behavior

The extension repeatedly opens the plugin settings page and the refresh obligation does not settle.

The Business plugin page contains the current tool/action schemas, but no `Refresh` control is exposed.

## Likely cause

`refreshManagedPlugin()` appears to require a verified `Refresh` button before it compares the installed tool schema with the expected schema.

Conceptually, the current order is:

```text
require Refresh control
→ compare installed schema with expected schema
→ if equal, mark current
```

This means an already-current connector cannot be marked current when the ChatGPT UI does not expose a Refresh button.

The relevant failure path appears to be:

```text
Exact connector settings or Refresh control could not be verified
```

Because a pre-claim failure remains retryable, maintenance can reopen the same settings page repeatedly.

## Expected behavior

The refresh workflow should first compare the installed schema with the expected schema.

Suggested behavior:

```text
if schema already matches:
[...]

PR #91, opened 2026-09-06
2.0.7: plus = gpt 5.6; pro = astra

totec448-spec:
## Result
This single snapshot commit contains the final 2.0.7 file changes on top of main; intermediate development commits are excluded. It integrates the browser ownership and native input fixes. Browser handoffs keep one elected document across navigation and worker suspension; Goal/Loop and finish requests retain exact turn ownership, retry temporary API failures, and deliver generated follow-ups through the next tool response. Provider destination, model and reasoning remain consistent across settings changes.

Native file workflows include bounded opt-in artifact downloads and image handling. Connector refresh now handles validated native cards with no Refresh action, while the companion popup requires confirmed protocol compatibility. Usage accounting recognizes the observed Sol model alias and uses current configurable cached-input rates. macOS notification actions receive the required alert-style packaging declaration.

## Validation
- Complete local integration suite: 3155 passed, 23 skipped before the final focused fixes; isolated shutdown tests included.
- Final affected suites: 177 passed; canonical artifact fixture follow-up: 16 passed. Finish test timer isolation also has a failing-before/passing-after regression. TypeScript and staged-diff checks pass.
- Shared-tree finish/exec regression rerun: 163 passed, 3 skipped.
- Public snapshot privacy checks pass. Maintainer logs and local session evidence are excluded.
- Windows/macOS/Linux CI all pass against snapshot 7c38fcb (run 34140556583). Local Windows installation completed; installed main/preload/renderer bytes match a fresh build of the final runtime.

Live Windows testing confirms ordered, exactly-once queued-stage delivery and one active worker. Model compliance with completing every stage before calling finish remains distinct from transport correctness. Actual packaged macOS notification/focus behavior and native Linux behavior are not claimed as locally verified.

PR #93 startup/reopen behavior is covered. Its mandatory residency and Dock-removal proposal is deferred because it changes existing close preferences. Broader product proposals remain separately scoped.

PR #92, opened 2026-09-06
Fix plugin refresh loop when Refresh control is unavailable

TaeyanG4:
## Summary

Fix the ChatGPT Business plugin-refresh loop reported in #90 when the workspace either already has the current schema or cannot refresh a stale schema in place because the settings UI exposes no `Refresh` button.

## Root cause

`refreshManagedPlugin()` previously required a visible/enabled `Refresh` control before comparing the installed tool schema with the expected schema. On ChatGPT Business, the plugin details page can expose the current action schemas without exposing `Refresh`, so an already-current connector failed before it could be acknowledged as current.

A second live-repro case showed that when the schemas genuinely differ and Business exposes no `Refresh` control, the request stayed retryable (`attempted: false`). Extension maintenance then reopened the same plugin settings page on every retry.

## Change

- Compare the installed schema with the expected schema before requiring `Refresh`.
- If the schemas already match, acknowledge the publication as current without clicking anything.
- Only require `Refresh` when the schemas actually differ.
- If the schemas differ and no `Refresh` control exists, record a durable manual-action state instead of leaving the request retryable.
- Exclude that manual-action state from automatic browser pickup so the settings page is not reopened indefinitely.
- A later local schema change creates a fresh refresh obligation, so the manual state does not permanently suppress future updates.
- Add regression coverage for both no-Refresh cases and the bridge protocol path.

## Validation

- `npm test -- --run test/plugin-refresh-workflow.test.ts` — 11/11 passing.
- `npm test -- --run test/plugin-refresh.test.ts` — 18/18 passing.
- `npm test -- --run test/input-delivery-integration.test.ts` — 24/24 passing.
- `npm run typecheck` — passing.
- `git diff --check` — passing.
- Earlier full `npm run verify` reached 2921 passing tests but hit one unrelated Windows/PowerShell locale-sensitive failure in `test/mcp.test.ts` (`scopes parser recovery to its failed batch command after an earlier mutation`); the plugin-refresh tests passed in that run.
- Live ChatGPT Business validation for the second (stale schema + no Refresh) case is pending before marking this PR ready for review.

Addresses #90.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -2264,12 +2264,12 @@ const HANDLERS = {
     if (!ownsDocument(source) || !/^[a-f0-9-]{36}$/i.test(String(message.id || ''))) return { ok: false };
     const tab = await chrome.tabs.get(source.tab);
     if (!ownsDocument(source) || pluginRefreshMarker(tab) !== message.id) return { ok: false };
-    if (!['claim', 'current', 'complete', 'fail'].includes(message.action)) return { ok: false };
+    if (!['claim', 'current', 'manual', 'complete', 'fail'].includes(message.action)) return { ok: false };
     const body = JSON.stringify({ action: message.action, id: message.id, appId: message.appId, connectorName: message.connectorName, tools: message.tools, versionId: message.versionId, error: message.error });
     if (body.length > 310000) return { ok: false };
     const result = await call('/plugin-refresh', { method: 'POST', body });
     if (!ownsDocument(source) || pluginRefreshMarker(await chrome.tabs.get(source.tab)) !== message.id) return { ok: false };
-    if (['current', 'complete', 'fail'].includes(message.action) && result.ok && result.data?.ok) void maintain();
+    if (['current', 'manual', 'complete', 'fail'].includes(message.action) && result.ok && result.data?.ok) void maintain();
     return result;
   },
   async model_catalog(message, _sender, source) {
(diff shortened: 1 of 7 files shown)


totec448-spec:

Both no-Refresh cases are incorporated: an exact current schema settles without a Refresh button, while a stale schema records durable manual action. The native settings path is also covered. Automatic refresh remains off by default.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/plugin-refresh.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/plugin-refresh-dom.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

totec448-spec:
@TaeyanG4 I incorporated your missing-Refresh-control fixes into #91 but failed to preserve your commit attribution. I am sorry. [CONTRIBUTORS.md](https://github.com/totec448-spec/chat-on-steroids/blob/main/CONTRIBUTORS.md) now credits #92, and [86d7014](https://github.com/totec448-spec/chat-on-steroids/commit/86d7014a2bc18b6dbdf61972ec275e2fc1eb3b40) adds retroactive co-author credit linked to your GitHub account. Thank you for tracking down both cases.

PR #93, opened 2026-09-07
Fix macOS tray app window lifecycle

ahrorbeksoft:
## Summary

- keep macOS running as a true menu-bar accessory with no Dock/App Switcher presence
- open the control panel on every primary app launch and reopen it from tray/native activation
- hide the control panel to the tray/menu bar on close instead of quitting
- pin the legacy close-to-tray setting to the new always-resident behavior
- verify `LSUIElement` in macOS package smoke checks and update lifecycle/docs coverage

## Validation

- `npm run typecheck`
- `npx vitest run test/window-lifecycle.test.ts test/packaging.test.ts test/renderer-state.test.ts` — 54 passed
- `npm test` — 2,844 passed, 100 skipped
- `node scripts/smoke-macos-bundle.mjs arm64`
- packaged `Info.plist` verified with `LSUIElement=true`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -491,7 +491,7 @@ single-instance lock
   → if feature is off: pause live execution, preserve history, persist the safe projection
   → restore continuations AFTER swarm, because recovery may need to repair prime ownership
   → install renderer CSP + deny browser permissions
-  → register fixed IPC → enable the native window activation gate → create/show window + tray
+  → register fixed IPC → enable native activation → macOS accessory policy → create tray + show control panel
   → queue legacy attribution repair asynchronously
   → start bridge if recording OR multi-agent
   → start session-retention maintenance independently of recording admission
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ ChatGPT is a good engineer trapped in a text box. Developer mode lets it call MC
 - **Plans, Goal and Loop.** Split a request into editable tasks or generate follow-ups through a separate ChatGPT helper or the API. Astra can receive the next task through `session_finish` in the same turn, without opening another model turn.
 - **You stay the permission boundary.** Only the folders you approve are visible. Each capability is a switch. Read-only mode is a single kill switch. Nothing runs on this machine that you did not turn on.
 
-It runs in the tray, hosts no model of its own, and works with the ChatGPT you already use in the browser.
+It runs in the tray, hosts no model of its own, and works with the ChatGPT you already use in the browser. On macOS it lives in the menu bar without occupying the Dock or App Switcher.
 
 ## Download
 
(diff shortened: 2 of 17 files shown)


ahrorbeksoft:
@codex review

Maximapple:
This is the one open PR here whose work has **not** been absorbed upstream — `LSUIElement` appears nowhere on `main`, and `src/main/index.ts` has no activation-policy or Dock handling. So it is worth rebasing rather than abandoning.

I test-merged the branch onto `main` (`067c40d`) to see what 2.0.8 actually broke. Five conflicts, and only one of them is a real decision — the rest are the merge misreading deletions.

**The trap: your base predates two settings that 2.0.8 added.** The merge base is `0f3ec75` (6 Sep), where the config read:

```ts
ui: { minimizeToTray: true, autoConnect: false, privacyScreenshots: false, theme: 'dark' },
```

`startAtLogin` and `autoRefreshPlugins` did not exist yet. Your branch removes `minimizeToTray` deliberately — but git sees your side as *also* lacking the two new keys, so resolving these hunks with "take theirs" silently deletes both features:

- **`src/main/config.ts`** — resolve to `ui: { autoConnect: false, startAtLogin: false, privacyScreenshots: false, theme: 'dark', autoRefreshPlugins: false }`. Drop only `minimizeToTray`.
- **`src/renderer/main.ts`** — both hunks: keep the `startAtLogin` lines, drop only the `minimizeToTray` ones.

**Two that are simply "keep both":**

- **`electron-builder.yml`** — `main` added `NSUserNotificationAlertStyle: alert` under the same key. Keep it *and* add your `LSUIElement: true`.
- **`src/main/index.ts`** — `main` added `if (!isBackgroundLaunch(process.argv)) windowActivation.request();` above the comment you reworded. Keep that line and take your comment. It also complements the PR rather than fighting it: a background launch is exactly the case where the panel should *not* open, which your "open on every primary launch" goal already implies.

**The one real decision — `test/window-lifecycle.test.ts`:**

```
main : expect(operations.splice(0)).toEqual(['maximize', 'show', 'focus']);
yours: expect(operations.splice(0)).toEqual(['show', 'maximize', 'focus']);
```

2.0.8 changed the order to maximize-before-show. That is a behavioural change in the code you are touching, not a test detail, so it needs a look rather than a pick — my read is that maximizing before showing avoids a visible resize flash on first paint, which would be a deliberate improvement worth keeping. Your added block below it (the null-window re-create path asserting `createWindow` is called once) is new coverage that does not exist on `main` and should survive; it just needs the same order as whichever you keep.

None of that touches the substance of the PR. Worth noting for reviewers that this is a rebase away from mergeable, and that the accessory-app behaviour it implements is still missing on `main`.

(Unrelated data point while I was in here: #47 no longer needs rebasing — its `/commands/revivals/pending` reconciliation shipped in 2.0.7 (`b6cdb61`), both the bridge route and the extension-side call. Commented there separately.)

totec448-spec:
Thanks for working on the tray behavior. I am closing this version because it also removes the close-to-tray preference and changes window startup behavior across platforms. The concrete reopen-size bug is fixed in #134 through the smaller change from #122. If there is still a distinct macOS tray failure, a focused reproduction would make that much easier to address without changing everyone's window policy.

PR #95, opened 2026-09-07
feat: support custom OpenAI-compatible provider for Goal/Loop

lookvincent:
Lets the api backend run on a user-provided OpenAI-compatible endpoint (Ollama, vLLM, LM Studio, or a gateway) instead of only OpenRouter.

Adds goal.provider {kind: openrouter|custom, baseUrl} (OpenRouter stays the default; old configs parse unchanged) plus a separate customProviderApiKey secret slot. Custom requests omit OpenRouter-only fields and use the OpenAI-compatible top-level reasoning_effort field when configured; keyless local servers send no Authorization header. Invalid base URLs fail settled, and both chat completions and model catalogue requests refuse redirects.

Provider selection is frozen per request: endpoint, model and reasoning remain consistent while the matching provider-kind key is read, so a provider switch cannot send one provider credential to another endpoint. Credential changes retire drafts only for the active provider.

Verification: deterministic provider-switch regression plus Goal and IPC coverage; 160 focused tests pass, along with typecheck and the privacy gate. The full suite passed 2852 tests; its only two failures are existing local-environment assertions (Homebrew rg shadowing the bundled binary, and the development tunnel resource fallback).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -338,7 +338,7 @@ src/shared/types.ts           config/app/IPC types and Capabilities
 
 ── browser ────────────────────────────────────────────────────────────────
 src/main/bridge.ts            extension HTTP bridge + compaction/worker orchestration
-src/main/goal.ts              Goal/Loop OpenRouter driver, durable obligations, one draft per turn
+src/main/goal.ts              Goal/Loop LLM driver (OpenRouter default, custom endpoint optional), durable obligations, one draft per turn
 src/main/agents.ts            the one global star-topology multi-agent broker
 extension/manifest.json       MV3 composition root: service worker, isolated scripts/CSS, MAIN-world Fiber, popup, host/extension permissions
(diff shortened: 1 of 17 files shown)


lookvincent:
Design intent is written up in #98 (including how this relates to the backend split).

totec448-spec:

Custom OpenAI-compatible Goal providers are incorporated, with separate credentials, keyless support, redirect refusal and one provider/model/reasoning snapshot across deferred request preparation.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/src/main/goal.ts) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/goal.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

PR #102, opened 2026-09-07
Match the ChatGPT model picker structurally so localized UIs work (fixes empty model selector on pl-PL)

piotrczukwinski:
## Root cause

Model discovery always failed with `picker_unavailable` on non-English ChatGPT interfaces (observed on `pl-PL`): every element of the composer's model picker was located by **English-only text** in `extension/chatgpt-dom.js`:

| Element | Old English-only match | Polish UI reality |
| --- | --- | --- |
| Picker trigger button | caption regex `(?:Instant\|Medium\|High\|Pro\|Thinking effort)` | caption is the localized effort name, e.g. „Bardzo wysoki" |
| Power row | `aria-label="Power"` | „Moc" |
| Select-model toggle | `aria-label="Select model"` | „Wybierz model" |
| Effort level | announcement `"<label>, N of M"` | „Pro, 5 z 5." |
| Latest routing row | literal `Latest` | „Najnowsze" |
| SEND/SPEECH anchors | `aria-label^="Send"/"Dictate"/"Voice"` (fallbacks) | „Wyślij", „Rozpocznij dyktowanie", „Uruchom tryb głosowy" |

So on any localized account the extension could not even open the picker, discovery returned zero models, and the app composer showed "No observed choices" / an empty model selector.

## The smallest change that fixes it

Element finding is now **structural first**, with known localized strings (en-US, pl-PL) only corroborating the structural find:

- **Trigger**: the only non-plus `button[aria-haspopup="menu"]` of the composer's trailing controls — its caption is no longer load-bearing. (Also fixes discovery on layouts where the trailing testid container is absent: the trigger now falls back to `#prompt-textarea`'s form.)
- **Power row**: the picker `menuitem` that owns a `role="slider"`.
- **Select-model toggle**: the picker `menuitem` without a slider.
- **Effort decoding**: `"<label>, N <of> M"` with per-locale label sets and the count word ("of"/"z").
- **Latest label, temporary-chat controls, SEND/SPEECH anchor fallbacks**: accept the observed localized strings alongside English.

Per-locale tables cover `en` and `pl` today; other locales fall back to English strings, and the structural anchors alone still locate every picker element on any language.

Discovery ordinal logic, restore proof, and fail-closed admission in `inspectModelSettings`/`selectModelSettings` are **unchanged** — only element finding and label decoding changed.

## How it was validated

End-to-end on a `pl-PL` ChatGPT account (Windows, unpacked extension):

1. Loaded the modified extension (reload via `chrome://extensions`), reloaded the ChatGPT tabs.
2. Triggered model discovery from the app (Setup/composer "Reload ChatGPT models").
3. App log: `model discovery observed … models=3 … error=none` (previously `models=0 error=picker_unavailable` every attempt).
4. App composer now selects model and effort (`GPT-5.6 Sol · High`), model dropdown lists the account's models, power slider shows the discovered effort steps.
5. `node --check extension/chatgpt-dom.js`, `npm run typecheck`, and `npm run verify:privacy` all pass.

The picker was inspected live in both locales (en-US and pl-PL) before the change; all structural anchors (`data-testid="composer-intelligence-picker-content"`, `role="menuitem"`/`menuitemradio`, the slider, `aria-describedby` KeyboardAnnouncement) are identical across locales — only strings differ.

### DIFF
--- a/extension/chatgpt-dom.js
+++ b/extension/chatgpt-dom.js
@@ -49,7 +49,7 @@ var CLF_DOM = (() => {
   const STOP =
     'button[data-testid="stop-button"], button[data-testid="composer-stop-button"], ' +
     'button[aria-label="Stop streaming"], button[aria-label="Stop generating"], button[aria-label="Stop answering"]';
-  const SEND = 'button[data-testid="send-button"], form button[aria-label^="Send" i]';
+  const SEND = 'button[data-testid="send-button"], form button[aria-label^="Send" i], form button[aria-label^="Wyślij" i]';
   /** The composer's own trailing controls, where the send and dictation buttons live. */
   const TRAILING =
     '[data-testid="composer-trailing-actions"], [data-testid="composer-footer-actions"], ' +
@@ -57,7 +57,8 @@ var CLF_DOM = (() => {
   const SPEECH =
     'button[data-testid="composer-speech-button"], button[data-testid="composer-dictate-button"], ' +
     'button[aria-label^="Dictate" i], button[aria-label^="Voice" i], ' +
-    'button[aria-label="Start dictation" i], button[aria-label="Start Voice" i]';
+    'button[aria-label="Start dictation" i], button[aria-label="Start Voice" i], ' +
+    'button[aria-label^="Rozpocznij dyktowanie" i], button[aria-label^="Uruchom tryb głosowy" i]';
   const safe = (fn, fallback) => {
     try {
       const value = fn();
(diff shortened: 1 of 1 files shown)


totec448-spec:

The model-picker discovery repair is covered by the merged structural picker and scoped MAIN-world metadata path. This supersedes the English-label-dependent discovery path without requiring the original locale-table patch.

[Current source](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/extension/chatgpt-dom.js) · [Regression coverage](https://github.com/totec448-spec/chat-on-steroids/blob/b6cdb61224c6673c29c53d68d7d76f33abbc123f/test/chatgpt-dom-input.test.ts). Thank you for the contribution. This source disposition does not claim additional platform/account validation.

Issue #108, opened 2026-09-08
Normal Pro conversations: delayed wfr request identity causes WORKER_IDENTITY_LOST and late tool attribution


ventianima-lab:
## Environment
Windows, Chrome, Chat On Steroids 2.0.7 based on b6cdb61224c6673c29c53d68d7d76f33abbc123f; native picker verified gpt-6-pro/pro. A local external-controller patch is present; no identity gate was weakened.

## Observed
In ordinary browser Pro conversations, agents status fails with WORKER_IDENTITY_LOST although Core reads can proceed after attribution waits. The initially observed page UUID differs from the wfr_ request ID carried by MCP calls. Exact wfr identity arrives much later and previous records are retrospectively attributed.

One sanitized timeline:
- 18:18:19.709 UTC: initial page UUID attributed.
- Incoming tools carry a different wfr_ ID; reads wait about 15 seconds and agents calls about 30 seconds.
- 18:19:18.126: app requests attribution-recovery reload.
- 18:21:09.564: matching wfr_ evidence arrives.
A following turn showed about 54 seconds of delay; another conversation showed 194.446 seconds from delivery ACK to exact identity observation. A separate minimal Pro canary also had two agents-status identity rejections and created no worker.

## Diagnosis boundary
The extension already inspects metadata.request_id across turn messages, including non-api_tool messages. No proven local parser defect or trustworthy alternate correlation key was found. Delayed provider materialization is a hypothesis, not a confirmed root cause. Retrospective attribution is not proof that identity was available at call time. Later successful reads do not prove successful agent operations.

Likely investigation points: extension/fiber.js request metadata observation, src/main/session/correlation.ts, src/main/mcp/inbound.ts, and agents admission in src/main/mcp/kernel.ts.

## Expected
Expose delayed/unresolved caller identity distinctly and recover exact attribution without guessing from timing, focused tab, selected model, or the sole active conversation. Preserve the original operation and avoid duplicate submissions. Please investigate why normal Pro turns cannot obtain usable identity within the admission window; simply increasing waits may exceed connector deadlines.

Related #96 concerns scheduled runs with CALLER_IDENTITY_REQUIRED; this report is normal browser Pro turns with delayed wfr identity and WORKER_IDENTITY_LOST, so the trigger differs.

Private conversation IDs, workspace contents, raw logs, and credentials are omitted. Existing affected prompts were not replayed.

Maximapple:
[...]

PR #115, opened 2026-09-08
Spend the caller-evidence window once per request, not once per call

Maximapple:
From a real run's log, not a hypothesis. The interesting lines:

```
15:02:12  multi-agent: worker-34 passed the 400000 token ceiling; it keeps working but cannot be woken again
15:16:30  tool write_stdin rejected in 15002 ms: session 89859 is not proven to belong to this durable session
15:16:30  request POST mcp/core → 200 in 30015ms
15:16:50  warn request attribution: no page evidence for wfr_01a0…780a within 20000ms
15:17:09  tool exec_command rejected in 15002 ms: CALLER_IDENTITY_REQUIRED
15:17:09  request POST mcp/core → 200 in 30010ms
```

Same request id on both. **30 seconds per tool call**, one after another, against a chat whose
page could no longer report who it was.

### Why the wait exists, and where it is charged wrongly

`awaitRequestCorrelation` waits because a call can arrive before the page evidence that names
its conversation — the case it was written for was about eight seconds late. That is sound.

What it is not is a per-call cost. Every tool call of one ChatGPT turn carries the same request
id, so once the page stops reporting, **each call spends the whole window again for the same
answer**. A handful of those is long enough for ChatGPT to abandon the turn with "Message
delivery timed out. Please try again."

The calls in that log were going to fail either way; the evidence was genuinely gone. It was the
waiting that ended the turn.

### Change

The window is spent once per request id. A later call in the same turn gets the answer
immediately instead of repeating it.

Nothing else moves, and the ordering is what keeps that true:

- the immediate read of an already-known owner runs **first**, so evidence that lands late is
  still returned normally;
- a different turn has spent nothing and waits in full;
- a correlation that does arrive clears the mark, so nothing stays remembered as hopeless once
  it is not;
- the set is bounded.

### Verification

`test/correlation.test.ts` — *waits the evidence window once per request, not once per call* —
asserts all four: the first call pays the window, a second call in the same turn does not, an
unrelated turn still does, and late evidence is answered.

Against the previous behaviour it fails with
`a later call in the same turn must not pay it again: expected 67 to be less than 30`.

`npm run typecheck` clean. 3161 passing. The three failures left are environmental and reproduce
identically on a pristine `main` checkout here, which has one more.

### Scope

This does not fix the chat in that log — a worker past its context ceiling has a real problem
that this cannot solve. It stops that problem from costing 30 seconds a call while the turn is
still trying to work, which is what turned a stalled worker into an abandoned turn.

Same symptom as #110, different route to it: there a request in flight was destroyed, here every
request pays for evidence that is not coming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,21 @@ the app refuses the extension and asks you to reload the matching copy.
   applied there to a caption, and not to the action that kills live work. A client that really
   is unready is still replaced, one interval later.
 
+## Unreleased
+
+### Fixed
+- **A chat whose page has stopped reporting no longer spends half a minute per tool call finding
+  that out.** Every tool call of one ChatGPT turn carries the same request id, and the wait for
+  the page evidence that names the caller was charged to each of them separately. Against a chat
+  that could no longer report — a worker past its context ceiling, in the run this came from —
+  that was 30 seconds per call, repeatedly. A handful of those is long enough for ChatGPT to give
+  up on the turn with "Message delivery timed out", so the waiting ended the turn rather than the
+  missing evidence, which was going to fail those calls in any case.
+
+  The window is now spent once per request id. Nothing else changes: the first call still waits
+  it in full, a different turn is unaffected, and evidence that arrives late is still answered
+  immediately, because the already-known answer is read before any of this is consulted.
+
 ## [2.0.7] — 2026-09-07
 
 **plus = gpt 5.6; pro = astra**
--- a/src/main/session/correlation.ts
+++ b/src/main/session/correlation.ts
@@ -51,6 +51,33 @@ const CORRELATIONS_STATE_VERSION = 5;
 
 const byRequest = new Map<string, RequestCorrelation>();
 const waiters = new Map<string, Set<() => void>>();
+
+/**
+ * Requests whose evidence window has already gone by unanswered.
+ *
+ * The wait exists because a call can reach this app before the page that proves who made it —
+ * measured at about eight seconds late in the case it was written for. What it is not is a
+ * per-call cost: every tool call of one ChatGPT turn carries the same request id, so a page
+ * that has stopped answering makes each of them spend the whole window again. On 2026-09-08
+ * that was 30 seconds per call against a chat whose page could no longer report, and a handful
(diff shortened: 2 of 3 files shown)


totec448-spec:
🤞🤞🤞🤞

totec448-spec:
Time to let gpt 6 pro with 2 gpt 5.6 pro sub agents improve the backend handling. Enough ui for now. Good pr

totec448-spec:
Thanks for catching the repeated wait. I brought the fix into #134 with one adjustment: each caller now gets its remaining time from the same starting point, so a short kernel wait does not consume the recorder's longer allowance. The regression tests and all three CI platforms pass. Closing this PR as incorporated; #108 stays open because missing provider identity is a separate problem.

Maximapple:
Superseded by #134, and by a better version of the same idea — closing.

This PR spent the evidence window once per request id using a `Set` of ids that had already
waited it out. 2.0.8 keeps a `Map` of when each window *started* instead:

```ts
const evidenceWindowStarts = new Map<string, number>();
```

> Keep the start rather than a spent flag: the recorder's longer grace must remain available
> after a shorter identity lookup expires.

That is the case this PR got wrong. Callers do not share one deadline — a short identity lookup
and the recorder's longer grace both consult this, and a binary "spent" flag lets the shorter one
retire the window for the longer one. Measuring from the start gives every caller the remainder
of its own allowance, which fixes the same 30-seconds-per-call waste without that side effect.

Confirmed while carrying the browser-tool branch onto 2.0.8: upstream's implementation replaces
this one cleanly, and this branch's tests came out with it.

One thing worth passing back rather than losing, from resolving that merge. The reconciliation in
the new `correlation.ts` reads sessions through `listAllSessions()`:

```ts
sessions = await listAllSessions();
```

`store.ts` documents that wrapper as the bounded view, with its own comment saying never to use it
for retention or identity — past `MAX_LISTED_SESSIONS` it answers "no such session" for a session
that exists. This is the code that decides which conversation owns a request id, so on an install
past the cap it can hand a request to the wrong chat. The same file already names
"request-correlation crash-window reconciliation" as one of the callers that must use
`readEverySummary()` instead.

Not filing it as a separate issue since it is one import in code that just landed, but it is a
real edge and worth a look.

PR #122, opened 2026-09-09
Preserve window size when reopening

### LINKED ISSUE #121: Minimized window reopens maximized instead of preserving its size
## Bug

When the main window is resized, minimized, and then reopened (for example from the macOS Dock), it comes back maximized instead of at the size the user left it.

## Root cause

`showWindow()` first restores a minimized `BrowserWindow` and then unconditionally calls `maximize()` whenever the window is not fullscreen. The same `showWindow()` helper is shared by initial presentation and later native/tray/second-instance activation, so the startup maximization policy is also applied to user-driven reopen.

## Expected behavior

Keep the existing maximized presentation for a newly created window, but do not change the geometry of an existing window when it is restored or shown again.

## Proposed fix

Move the startup-only maximization into the `ready-to-show` path and leave `showWindow()` responsible only for restore/show/focus. Add a deterministic regression test covering both startup and minimized reopen.

## Validation performed locally

- focused native window lifecycle tests pass
- TypeScript typecheck passes
- production build passes
- packaged macOS arm64 bundle/native runtime smoke passes
- installed macOS build manually verified: resized window preserves its size after minimize and Dock reopen

becoolmin:
Fixes #121.

## What changed

`showWindow()` was applying the startup maximization policy every time an existing window was reopened. A minimized window was restored and then immediately maximized, discarding the user's previous size.

This keeps the existing native-maximized presentation for a newly created window by moving `maximize()` into the `ready-to-show` path, while `showWindow()` now only restores, shows, and focuses an existing window. The window lifecycle regression test covers both initial presentation and minimized reopen.

## Validation

- [ ] `npm run verify` passes. Local verification reaches the test suite but has two environment-dependent baseline failures that reproduce unchanged on clean `upstream/main`: the machine-wide `/opt/homebrew/bin/rg` wins the bundled-ripgrep assertion, and a locally available tunnel client is discovered by the explicit-binary fallback test. The window change is not involved in either failure.

Additional results:

- `npm test -- --run test/window-lifecycle.test.ts`: 12/12 passed
- `npm run typecheck`: passed
- `npm run verify:privacy`: passed
- `npm run dist:dir:mac:arm64`: passed
- packaged darwin-arm64 native/runtime smoke: passed
- macOS arm64 bundle audit: passed
- manual installed-app check: a resized window retains its size after minimize and Dock reopen

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -136,7 +136,13 @@ function createWindow(): void {
   window.once('ready-to-show', () => {
     // A renderer can finish loading after Cmd+Q has already entered bounded teardown. Never let
     // that late native event make the app visible again while `will-quit` is draining.
-    if (!quitting) showWindow();
+    if (!quitting) {
+      // Newly created windows intentionally start maximized. Keep that startup-only presentation
+      // here so later tray/Dock/native activation can show an existing user-sized window without
+      // overwriting its geometry.
+      if (!window?.isFullScreen()) window?.maximize();
+      showWindow();
+    }
   });
 
   // A renderer that fails to load leaves a blank window with no other clue, so
@@ -200,9 +206,6 @@ function showWindow(): void {
     return;
   }
   if (window.isMinimized()) window.restore();
-  // Apply maximization before showing the window so startup has the native maximized
-  // frame from its first visible paint. Preserve a user's explicit F11 fullscreen choice.
-  if (!window.isFullScreen()) window.maximize();
   window.show();
   window.focus();
 }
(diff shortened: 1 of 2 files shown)


totec448-spec:
Thanks, this was a useful, focused fix. It is now included in #134: the initial window can still maximize at startup, but reopening preserves the user's window state. The lifecycle regression test and all three CI platforms pass. Closing this PR as incorporated.

totec448-spec:
@becoolmin Your window-reopen fix was incorporated into #134, but I failed to preserve your commit attribution. I am sorry. I have added explicit credit for #122 to [CONTRIBUTORS.md](https://github.com/totec448-spec/chat-on-steroids/blob/main/CONTRIBUTORS.md) and retroactive co-author credit in [86d7014](https://github.com/totec448-spec/chat-on-steroids/commit/86d7014a2bc18b6dbdf61972ec275e2fc1eb3b40). GitHub now links that credit to your account. Thank you for the fix.

PR #127, opened 2026-09-09
Fix Compact & Resume recovery loop on blocked handoff

### LINKED ISSUE #126: Compact & Resume can loop on restored composer draft after pre-Send handoff rejection
## Summary

Automatic **Compact & Resume** can enter a repeated recovery/reload loop when ChatGPT restores a stale or otherwise occupied composer draft and the COS handoff prompt is rejected before Send.

Observed failure sequence on 2.0.8:

1. A long-running chat crosses the auto-compaction threshold (observed context ~4.7M against a 400K model window).
2. COS files an automatic compaction ticket.
3. The source page tries to inject `[[CLF-HANDOFF:...]]`.
4. ChatGPT refuses insertion because the composer is already occupied/restored.
5. The automatic ticket remains open because no source Send checkpoint was crossed.
6. The pickup watchdog reloads the chat to recover the uncollected ticket.
7. ChatGPT restores the same composer draft, so the same injection failure repeats.

The visible symptoms include repeated:

> Reloaded chat to recover an uncollected compaction ticket

followed by:

> ChatGPT would not accept the handoff instruction — clear the message box and try again.

The existing watchdog is bounded per ticket, but without a turn-local refusal fence the same over-threshold working turn can immediately file another automatic ticket after a terminal pre-Send refusal.

## Root cause

There are two related state-machine gaps:

- **Page side:** a persistent occupied composer was treated like a transient DOM failure, so an automatic ticket stayed recoverable even when the page could prove the prompt never reached Send.
- **Bridge side:** after terminally abandoning such a ticket, later activity from the same working turn could immediately file a fresh automatic ticket because the threshold condition is level-triggered.

A stale COS-owned handoff/resume draft may also survive reload and block the next ticket even though it is safe to identify by its CLF marker + canonical COS instruction.

## Proposed fix

The attached PR will:

- recognize and clear **only a proven stale COS-owned** `CLF-HANDOFF` / `CLF-RESUME` draft with a different token plus canonical COS text;
- preserve ordinary user drafts and same-token edits;
- add a `sourceLost` path that durably aborts an automatic source handoff only while the WAL still proves no Send happened (`not-attempted` / `attempted-unresolved`);
- refuse that abort after `sourceDispatch`, preserving at-most-once semantics;
- add a turn-local fence so the same working turn cannot immediately re-file another auto-compaction after this terminal refusal;
- clear that fence on a genuine turn boundary;
- keep transient missing/replaced-composer failures on the existing recoverable path.

## Regression coverage

The patch adds tests for:

- persistent user draft: retire the automatic ticket without modifying the draft;
- stale COS draft: clear it and allow the current ticket to proceed;
- transient missing composer: keep the ticket recoverable;
- durable source-side pre-Send abort and rejection after dispatch;
- no immediate automatic re-file in the same working turn;
- normal re-arming after a genuine turn boundary;
- background forwarding of the `sourceLost` checkpoint.

## Validation performed

On macOS arm64 with the 2.0.8 source tree:

- `npm run typecheck` — PASS
- focused compaction/content/continuation/extension suites — PASS
- full `npm test` — **3335 passed, 106 skipped, 0 failed**
- production build — PASS
- packaged macOS arm64 bundle/runtime smoke checks — PASS

A locally packaged 2.0.8 build containing the patch has also been run against the affected installation.

frytufrytu:
Closes #126.

## What this fixes

Prevents automatic Compact & Resume from cycling through reload/recovery when ChatGPT restores an occupied or stale composer draft and the source handoff is rejected before Send.

The failure looked like:

- context crosses the auto-compaction threshold;
- COS files an automatic ticket;
- `[[CLF-HANDOFF:...]]` insertion is rejected because the composer already contains restored text;
- the ticket stays open;
- pickup reloads the same chat;
- ChatGPT restores the same draft;
- the same pre-Send rejection repeats.

The existing per-ticket watchdog is bounded, but because auto-compaction is level-triggered the same working turn could file another ticket after the previous one was abandoned.

## Changes

- **Content script:** distinguish a persistent occupied composer from transient DOM/composer loss.
  - Preserve ordinary user drafts.
  - Preserve same-token edits.
  - Clear only a proven stale COS-owned `CLF-HANDOFF` / `CLF-RESUME` draft with a different token and canonical COS text.
  - Report `sourceLost` when an automatic ticket is blocked before Send by a persistent other draft.

- **Continuation WAL:** add `abortContinuationSourceBeforeSendNow()`.
  - Allowed only while the source checkpoint still proves no Send happened (`not-attempted` / `attempted-unresolved`).
  - Refused after `sourceDispatch`, preserving at-most-once semantics.

- **Bridge:** handle `/compact { sourceLost: true }` durably.
  - Return retryable failure if the terminal WAL write cannot persist.
  - Clear the compaction watch/repair only after successful durable abort.

- **Auto-compaction scheduler:** fence the rejected automatic compaction for the current working turn.
  - More observations from that same turn cannot immediately file a new token.
  - A real turn boundary clears the fence and allows later compaction normally.

- **Extension worker:** forward the new `sourceLost` checkpoint.

## Safety properties

This intentionally does **not** broaden generic composer clearing or generic cancellation:

- user text remains untouched;
- transient missing/replaced composer failures remain recoverable;
- stale cleanup requires a COS marker, canonical COS instruction, and a different token;
- once `sourceDispatch` has been crossed, `sourceLost` is rejected rather than cancelling an ambiguous send.

## Tests

Added regressions for:

- persistent user draft -> durable pre-Send retirement without modifying the draft;
- transient missing composer -> ticket remains recoverable;
- stale COS draft -> clear stale text and reach the current send fence;
- durable source abort only before dispatch;
- no immediate auto-refile in the same working turn;
- re-arm after a genuine turn boundary;
- background forwarding of `sourceLost`.

## Validation

Validated on the 2.0.8 tree on macOS arm64:

- `npm run typecheck` ✅
- focused compaction/content/continuation/extension suites ✅
- `npm test` ✅ — **3335 passed, 106 skipped, 0 failed**
- production build ✅
- packaged macOS arm64 bundle/runtime smoke checks ✅

A locally packaged 2.0.8 build with this patch was also installed and run on the affected machine.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -2830,6 +2830,9 @@ const HANDLERS = {
         ...(typeof message.token === 'string' && message.sourceAttempt === true
           ? { token: message.token, sourceAttempt: true }
           : {}),
+        ...(typeof message.token === 'string' && message.sourceLost === true
+          ? { token: message.token, sourceLost: true }
+          : {}),
         ...(typeof message.token === 'string' && message.sourceDispatch === true
           ? { token: message.token, sourceDispatch: true }
           : {}),
--- a/extension/content.js
+++ b/extension/content.js
@@ -8032,19 +8032,29 @@
       CLF_DOM.conversationId() === forId;
     let attemptCrossed = false;
     const automaticTicket = job && job.automatic === true;
-    const abandonBeforeSend = async (why) => {
+    const abandonBeforeSend = async (why, retireAutomatic = false) => {
       if (!current()) return;
       nativeBusy = false;
       nativePhase = '';
       pressedAt = 0;
       localError = why;
-      // A page/DOM failure is not a verdict on an automatic ticket. Keep it on the
-      // continuation WAL so the app's next pickup reload can collect the same work. A manual
-      // press keeps its historical immediate-abort behaviour; the user is still present and
-      // can retry it without leaving an invisible job behind.
+      // A transient page/DOM failure is not a verdict on an automatic ticket. Keep it on the
+      // continuation WAL so the app's next pickup reload can collect the same work. A composer
+      // already holding another draft is different: ChatGPT restores that draft across reloads,
+      // so the caller can retire this pre-Send ticket instead of scheduling the same refusal.
+      // A manual press keeps its historical immediate-abort behaviour; the user is still present
+      // and can retry it without leaving an invisible job behind.
       if (!automaticTicket) {
         job = null;
         await ask({ type: 'compact', conversationId: forId, cancel: true }).catch(() => undefined);
+      } else if (retireAutomatic) {
+        // This is not the user-facing Cancel path. The bridge accepts sourceLost only while its
+        // durable checkpoint still proves no Send happened (`not-attempted` or
+        // `attempted-unresolved`). If another page crossed sourceDispatch meanwhile, this refuses
+        // and the ambiguous attempt remains alive rather than being cancelled underneath it.
+        const lost = await ask({ type: 'compact', conversationId: forId, token, sourceLost: true }).catch(() => null);
+        if (lost && lost.ok === true && lost.data && lost.data.aborted === true) job = null;
+        else localError = replyError(lost) || 'The blocked handoff could not be safely retired; it was not sent twice.';
       }
       if (!current()) return;
       renderControl();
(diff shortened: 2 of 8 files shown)


totec448-spec:
Thanks for tracking down the restored-draft loop. I incorporated the pre-Send retirement fix in #134, with two changes: drafts are always preserved, and the refusal survives an app restart for that same turn. The next turn can still compact normally. The regression tests and all three CI platforms pass, so I am closing this PR as incorporated.

totec448-spec:
@frytufrytu You deserve credit for identifying and fixing the blocked-handoff recovery loop. I adapted your work into #134 and failed to preserve your commit attribution. I am sorry. [CONTRIBUTORS.md](https://github.com/totec448-spec/chat-on-steroids/blob/main/CONTRIBUTORS.md) now credits #127 and explains the adaptation, and [86d7014](https://github.com/totec448-spec/chat-on-steroids/commit/86d7014a2bc18b6dbdf61972ec275e2fc1eb3b40) adds retroactive co-author credit linked to your GitHub account.

PR #131, opened 2026-09-09
Stop the Linux DEB smoke failing on its own teardown

### LINKED ISSUE #129: Release candidate: Linux DEB smoke fails on sandbox cleanup after the test itself passed
## What happened

`Package Linux x64` fails at the very end of **Launch installed DEB normally under Xvfb**, after the smoke test itself has already succeeded. The step's own assertions all pass — the app starts, the window loads, the renderer reports ready — and then the step exits 1 on directory cleanup:

```
[info] Chat On Steroids 2.0.8+<sha> starting on linux-x64
[info] app started
[info] renderer state ready
[info] window loaded
rm: cannot remove '/tmp/tmp.vFazdhiPWK/config/chat-on-steroids': Directory not empty
##[error]Process completed with exit code 1.
```

Second run, same job, different sandbox — same ending:

```
rm: cannot remove '/tmp/tmp.8LD6aePKeX/config': Directory not empty
```

Because `Assemble release candidate` has `needs: [package, sources]`, one such failure skips it, so a release candidate produces no assembled output and no `Chat-On-Steroids-Extension.zip` even though **every** package job succeeded.

## Root cause

The step creates a sandbox and registers an unconditional cleanup trap, then kills the app on a timer:

```bash
deb_smoke_root="$(mktemp -d)"
trap 'rm -rf "$deb_smoke_root"' EXIT
...
CLF_DEBUG=1 timeout --signal=TERM --kill-after=5s 45s xvfb-run -a /usr/bin/chat-on-steroids >deb-gui.log 2>&1
```

`timeout` signals `xvfb-run`, not the process tree beneath it. Electron's helper processes — and anything the app spawned itself — can still be writing into `$XDG_CONFIG_HOME` when the trap fires immediately after the greps. `rm -rf` then walks a directory that is gaining entries while it deletes, which is exactly the `Directory not empty` failure mode.

The step runs under `set -euo pipefail`, so that cleanup failure becomes the step's verdict. A sandbox teardown is deciding whether the smoke test passed.

## Why this matters beyond one flake

It is timing-dependent, so it is invisible until a runner is slow or the app has more to shut down — it failed twice consecutively here while passing on other runs of the same workflow with no application change between them. The comment already in this step records the same lesson from the other direction:

> 45s, not 12s. […] Four runs either side of it passed with no application change between them, so the budget was the variable. A GUI smoke that fails for being hurried teaches nobody anything.

The same applies to the teardown: a smoke test that fails for being hurried on the way *out* teaches nobody anything either — and unlike a hurried start, this one silently costs the whole release assembly.

## Suggested fix

Cleanup should not be able to fail the step:

```bash
trap 'rm -rf "$deb_smoke_root" 2>/dev/null || true' EXIT
```

If leaving the sandbox behind is undesirable, giving the process tree a moment first makes the delete succeed rather than merely be forgiven:

```bash
trap 'sleep 1; rm -rf "$deb_smoke_root" 2>/dev/null || rm -rf "$deb_smoke_root" 2>/dev/null || true' EXIT
```

The runner is ephemeral either way, so a residual temp directory has no cost.

Either version keeps every existing assertion intact — the exit-status check, the three required log lines, and the renderer-failure check all still decide the outcome. The only change is that a race in the teardown stops masquerading as a failed smoke test.

## Environment

Observed on `Release candidate` (`workflow_dispatch`), `ubuntu` runner, two consecutive runs. All other package jobs in both runs succeeded, macOS arm64 included.

Maximapple:
Fixes #129.

### The failure

`Package Linux x64` fails **after** its smoke test has passed. All three required log lines are present, no renderer error, and then:

```
[info] app started
[info] renderer state ready
[info] window loaded
rm: cannot remove '/tmp/tmp.vFazdhiPWK/config/chat-on-steroids': Directory not empty
##[error]Process completed with exit code 1.
```

Second run, same job, different sandbox, same ending (`/tmp/tmp.8LD6aePKeX/config`).

Because `Assemble release candidate` has `needs: [package, sources]`, that one job skips it — so a release candidate with **every** package job green produced no assembled output and no `Chat-On-Steroids-Extension.zip`.

### Why

```bash
deb_smoke_root="$(mktemp -d)"
trap 'rm -rf "$deb_smoke_root"' EXIT
...
CLF_DEBUG=1 timeout --signal=TERM --kill-after=5s 45s xvfb-run -a /usr/bin/chat-on-steroids
```

`timeout` signals `xvfb-run`, not the process tree beneath it. An Electron helper can still be writing into `$XDG_CONFIG_HOME` when the trap fires immediately after the greps, and `rm -rf` walking a directory that is gaining entries fails with `Directory not empty`. Under `set -euo pipefail` that cleanup becomes the step's verdict.

It is timing-dependent, which is why it stays invisible until a runner is slow or the app has more to shut down.

### The change

This workflow already solves exactly this problem, twenty lines further down in the AppImage smoke, and already states the rule:

```bash
# Cleanup is not product evidence. A process that created one last profile file
# after the smoke already proved renderer readiness must not turn the release red.
rm -rf "$target" 2>/dev/null || true
```

So this is not a new idea, just the same `cleanup_path_with_retries` helper applied to the one sandbox that never got it — deliberately the identical shape and name, so a reader meets one idiom rather than two.

Every existing assertion still decides the outcome: the `status -ne 124` check, the three required log lines, and the renderer-failure check are untouched. The only thing that changes is that a teardown race can no longer outrank them.

### Verified

`bash -n` clean on the extracted step body; indentation and block-scalar structure unchanged.

Behaviourally: the same branch that failed this job twice in a row now completes it. Before and after runs are linked in the comment below.

### Note on the comment style

The added comment avoids the em dash used elsewhere in the codebase, because this file is otherwise plain ASCII.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -219,7 +219,21 @@ jobs:
         run: |
           set -euo pipefail
           deb_smoke_root="$(mktemp -d)"
-          trap 'rm -rf "$deb_smoke_root"' EXIT
+          cleanup_path_with_retries() {
+            target="$1"
+            for attempt in 1 2 3 4 5; do
+              rm -rf "$target" 2>/dev/null && return 0
+              sleep 0.2
+            done
+            # Cleanup is not product evidence, the same rule the AppImage smoke below follows.
+            # `timeout` signals xvfb-run rather than the process tree beneath it, so an Electron
+            # helper can still be writing into the sandbox here, and `rm -rf` over a directory
+            # that is still gaining entries fails with "Directory not empty". Under `set -e`
+            # that teardown race outranked a smoke test whose own assertions had all passed,
+            # and skipped the release assembly with every package job green.
+            rm -rf "$target" 2>/dev/null || true
+          }
+          trap 'cleanup_path_with_retries "$deb_smoke_root"' EXIT
           mkdir -p \
             "$deb_smoke_root/home" \
             "$deb_smoke_root/config" \
--- a/test/packaging.test.ts
+++ b/test/packaging.test.ts
@@ -233,7 +233,11 @@ describe('cross-platform packaging targets', () => {
     expect(workflow).toContain("grep -Fxq 'Name=Chat On Steroids' \"$desktop\"");
     expect(workflow).toContain("grep -Fxq 'Icon=chat-on-steroids' \"$desktop\"");
     expect(debGui).toContain('deb_smoke_root="$(mktemp -d)"');
-    expect(debGui).toContain("trap 'rm -rf \"$deb_smoke_root\"' EXIT");
+    // Same shape the AppImage smoke below is held to: the teardown may retry, and may fail,
+    // but it may never decide the step. Only the assertions under it do that.
+    expect(debGui).toContain('cleanup_path_with_retries()');
+    expect(debGui).toContain("trap 'cleanup_path_with_retries \"$deb_smoke_root\"' EXIT");
+    expect(debGui).toContain('rm -rf "$target" 2>/dev/null || true');
     expect(debGui).toContain('HOME="$deb_smoke_root/home"');
     expect(debGui).toContain('XDG_CONFIG_HOME="$deb_smoke_root/config"');
     expect(debGui).toContain('XDG_CACHE_HOME="$deb_smoke_root/cache"');


Maximapple:
Behavioural proof, same branch, three consecutive runs of this workflow.

**Before** — `Package Linux x64` failed on the teardown while every other package job was green, so `Assemble release candidate` was skipped and the candidate produced no assembled output:

| run | Linux x64 | Assemble |
|---|---|---|
| [34344659069](https://github.com/Maximapple/chat-on-steroids/actions/runs/34344659069) | failure (`rm: cannot remove '/tmp/tmp.vFazdhiPWK/config/chat-on-steroids': Directory not empty`) | skipped |
| same run, `--failed` rerun | failure (`/tmp/tmp.8LD6aePKeX/config`) | skipped |

Both times the smoke's own assertions had passed first — `app started`, `renderer state ready`, `window loaded`, no renderer error.

**After** — [34349175536](https://github.com/Maximapple/chat-on-steroids/actions/runs/34349175536):

```
success  Package native library sources
success  Package Windows arm64
success  Package Linux x64
success  Package Windows x64
success  Package Linux arm64
success  Package macOS x64
success  Package macOS arm64
success  Assemble release candidate
```

No application change between the failing and passing runs other than this fix, and the assembled candidate exists again.

One note on the second commit, since it touches a test rather than the workflow: `test/packaging.test.ts` pinned the teardown by its exact old text, so it asserted precisely the shape that let cleanup decide the step. It is now pinned to the intent instead — a retrying helper, wired to the trap, whose last resort tolerates failure — which is what the same test already requires of the AppImage smoke a few lines below. That test is the only one that reads the workflow; full suite is otherwise unaffected.

PR #142, opened 2026-09-09
Add a browser tool: ChatGPT can drive a real web page

Maximapple:
Replaces #78, which you asked to be split. This is the browser capability on its own.

| | #78 | this |
|---|---|---|
| files | 156 | **20** |
| insertions | 31,773 | **3,746** |
| `docs/` | 45 files, 13,804 lines | **none** |
| native / session / scripts changes | yes | **none** |

Nothing here is QA history or implementation unrelated to driving a page. The native and session fixes that were mixed into #78 are being submitted separately — #139 and #141 are the first two.

### What it is

The desktop driver can click anywhere in a browser window but cannot see a web page: no refs, no DOM, no way to tell a link from the pixels around it. This tool speaks to the page instead, through the companion extension and CDP.

It lives in its own module, `src/main/mcp/tools-browser.ts`, rather than threaded through `tools-desktop.ts`. That is the main structural difference from #78 and it is deliberate: the three things you named are then reviewable in one place.

### Exact tab ownership

A run takes the newest ordinary tab or opens one. ChatGPT's own tabs are never driven; neither are browser pages or local files. The driven tab is placed in its own titled group so it is visibly driven, ungrouped when the run ends, and a group orphaned by an MV3 worker recycle is swept both before the next attach and on the extension's own wake timer — the case where nothing ever attaches again.

`test/browser-driver.test.ts` pins this, including that the sweep ungroups without ever closing the tabs.

### Live capability enforcement

`tabs`, `tabGroups` and host access are **optional** permissions, requested from the popup toggle rather than declared in the manifest. Chrome only prompts on a real user gesture, so the click that turns it on is the consent. Turning it off revokes them and drops any live session — off means the capability is gone, not merely unused. A refused request now says so in the popup instead of the switch snapping back silently.

The tool is registered only when control is exposed, and it is declared on the Desktop surface so the existing "never advertises a tool its surface does not declare" check covers it.

### Bounded output

Every action's reply is bounded, and the rendering is a plain function with its own tests (`test/browser-answer.test.ts`). It was inline before, reachable only through a live extension and a real browser, and it silently replaced every reply but `observe` and `status` with the word `ok` — discarding `hit`, `covered` and the driver build. That is the kind of gap a QA run found instead of a test.

### Discovery cost, stated

Desktop's worst-case `tools/list` goes from **7.9k to 12.9k bytes**. The ceiling in `test/mcp.test.ts` is raised deliberately to 13.5k and a per-tool ceiling for `browser` added at 5.5k, so the number stays visible and the next growth has to argue for itself rather than hiding under a surface total.

This is the honest price of the capability and the reason it is not on Core.

### One thing worth flagging

The worker imports the driver **statically**, because a service worker may not import dynamically — the specification forbids it and Chrome refuses at runtime with exactly that message. It was dynamic once, so the file parsed in the harnesses, every `browser_*` message failed in a real browser, and the tests stayed green throughout because they never execute one.

The three harnesses that evaluate `background.js` as a classic script now strip the import and supply the binding, and throw if the statement stops looking the way they expect — so the next person cannot quietly make the product dynamic again to keep a test parsing.

### Verified

`npm run typecheck` clean. Full suite **3,481 passed**; the four failures are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, which fail identically on unmodified `main` on this Windows machine and pass on Linux and macOS in CI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -350,12 +350,18 @@ jobs:
       - name: Package standalone extension
         run: |
           set -euo pipefail
-          (cd extension && zip -qr ../release/Chat-On-Steroids-Extension.zip . -x '*.map')
+          (cd extension && zip -qr ../release/Chat-On-Steroids-Extension.zip . -x '*.map' -x '*.d.ts')
           zip -q release/Chat-On-Steroids-Extension.zip LICENSE
           node -e "const m=require('./extension/manifest.json'),p=require('./package.json');if(m.version!==p.version)throw new Error('Extension '+m.version+' does not match app '+p.version)"
-          for required in manifest.json background.js chatgpt-dom.js content.js fiber.js overlay.css popup.html popup.css popup.js icons/icon16.png icons/icon32.png icons/icon48.png icons/icon128.png LICENSE; do
+          for required in manifest.json background.js browser-driver.js chatgpt-dom.js content.js fiber.js overlay.css popup.html popup.css popup.js icons/icon16.png icons/icon32.png icons/icon48.png icons/icon128.png LICENSE; do
             unzip -Z1 release/Chat-On-Steroids-Extension.zip | grep -Fxq "$required" || { echo "Extension zip missing $required" >&2; exit 1; }
           done
+          # And nothing that only exists for the build: a stray declaration file here means the
+          # published add-on ships something a browser cannot run.
+          if unzip -Z1 release/Chat-On-Steroids-Extension.zip | grep -q '\.d\.ts$'; then
+            echo "Extension zip carries TypeScript declarations" >&2
+            exit 1
+          fi
 
       - name: Create SHA-256 checksums
         run: |
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -37,6 +37,9 @@ extraResources:
     filter:
       - '**/*'
       - '!**/*.map'
+      # Type declarations describe the driver for this repository's own tooling. A browser
+      # has no use for them and a shipped extension should carry nothing it cannot run.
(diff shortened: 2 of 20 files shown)


Maximapple:
One follow-up deliberately left out of this PR, noted so it is not lost.

A connector keeps the `tools/list` it fetched when it was created, so a capability switched on afterwards adds a tool here that never appears there. QA hit exactly that with `browser`: the app was serving the tool and ChatGPT could not see it, with nothing to tell the two states apart.

The fix is a diagnostics check that compares what this server currently offers against the list the connector is actually holding, and says so on the same line. It is ~77 lines in `src/main/diagnostics.ts` and it only makes sense once there is a capability that can be switched on after a connector exists — so it belongs after this, not inside it.

Happy to send it as its own PR once this lands, or to fold it in now if you would rather review them together.

totec448-spec:
hey @Maximapple, I checked the production wiring as well as the tests. The proposed command collector/result settlement is not connected to a production bridge route, while the extension posts to a missing /browser/result endpoint. I’m keeping this unmerged until that complete path exists; isolated driver tests cannot prove a browser tool call can finish. Thanks for the prototype.

totec448-spec:
The current Desktop browser-control implementation already owns the browser tools and their command/result lifecycle, with dedicated engine, page, navigation and extension tests. This older parallel implementation would duplicate that authority. Closing it as superseded by the current src/main/browser-control.ts path, not as an unchanged merge of this branch.

Reviewed head: `4ac9a994d2ce`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #145, opened 2026-09-09
End an armed handoff the page proves ChatGPT never took

Maximapple:
One of the unrelated session fixes split out of #78.

### The gap

`abortContinuationSourceBeforeSendNow` ends the transaction from `not-attempted` and `attempted-unresolved`, and its comment says why it stops there:

> once the dispatch fence has been crossed the outcome is ambiguous and only ChatGPT's marker or an explicit user cancellation may end the transaction

That is right about **re-sending**, and nothing in this PR re-sends. But the ticket still has to end somewhere.

An armed dispatch the page then proved unaccepted falls past both pre-Send states, so `/compact` answers `409 source_send_not_releasable` and the ticket stands until `AUTOMATIC_HANDOVER_TTL_MS` — six hours — with `handoffAsked()` true, so no pickup can re-ask and the page will not retype it. Worse than the stall: the chat stays in `pendingAutomaticContinuations()` the whole time and `inspectSilentChats()` skips those, so a chat that is merely wedged also loses browser recovery for those six hours.

### Why the page's word is good enough here

Not a bare timeout. `send()` watches five independent acceptance signals — the composer clearing, a new conversation id, generation starting, the Stop control appearing, and a freshly rendered user message matching the text — for three seconds. The compaction flow has already stopped generation and waited for settle before it types, so a landed click flips generating/Stop within milliseconds. None of the five inside three seconds is the strongest negative this page can produce.

### The change

Only the refusal path is affected: the pre-Send abort still decides first, and just its `!aborted` answer now reaches `releaseContinuationSourceSendNow`, which aborts rather than re-arms.

The asymmetry with the destination half stays deliberate. The destination can prove nothing was sent — a chat with no id cannot have accepted a message — and may safely re-offer the brief; here the evidence is strong but not proof, so the transaction ends and nothing is offered again. Being wrong costs one abandoned compaction that auto-compaction refiles on the next qualifying turn, against six hours of a wedged chat.

### Verified

Fails before the change at the bridge level:

```
× ends an armed handoff the page proves ChatGPT never took
  AssertionError: expected 409 to be 200
```

Plus two unit tests: the armed dispatch is released and leaves the pending set, and a dispatch ChatGPT is *proven* to have taken (its marker bound) is still refused.

`npm run typecheck` clean. Full suite 3437 passed; the four failures are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, which fail identically on unmodified `main` on this Windows machine and pass in CI.

### DIFF
--- a/src/main/bridge.ts
+++ b/src/main/bridge.ts
@@ -144,6 +144,7 @@ import {
   abortContinuation,
   abortContinuationNow,
   abortContinuationSourceBeforeSendNow,
+  releaseContinuationSourceSendNow,
   attachSummary,
   beginContinuationDestinationSendNow,
   beginContinuationSourceSendNow,
@@ -2243,7 +2244,23 @@ async function handle(req: http.IncomingMessage, res: http.ServerResponse): Prom
           origin
         );
       }
-      if (!aborted) return json(res, 409, { error: 'source_send_not_releasable' }, origin);
+      if (!aborted) {
+        // The two pre-Send states are handled above. An armed dispatch that the page then
+        // proved unaccepted - none of send()'s five acceptance signals inside its window, on
+        // a chat this flow had already stopped and settled - falls past them, and refusing it
+        // leaves the ticket armed until the six-hour TTL with the chat kept out of browser
+        // recovery for all of it. Nothing is re-offered on either path; only the transaction
+        // ends.
+        const released = await releaseContinuationSourceSendNow(
+          checkpointToken,
+          'ChatGPT did not take the handoff instruction, and an armed dispatch is never sent twice'
+        );
+        if (!released) return json(res, 409, { error: 'source_send_not_releasable' }, origin);
+        compactionWatch.delete(id);
+        if (repairsInFlight.get(id)?.reason === 'compaction') repairsInFlight.delete(id);
+        changed();
+        return json(res, 200, { released: true, sessionId: entry.sessionId }, origin);
+      }
       compactionWatch.delete(id);
       if (repairsInFlight.get(id)?.reason === 'compaction') repairsInFlight.delete(id);
       changed();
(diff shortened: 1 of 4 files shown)


totec448-spec:
hey @Maximapple, I’m leaving this unmerged: an acceptance timeout after dispatch is still ambiguous, so it cannot prove that ChatGPT never took the summary request. Releasing that source could admit more work before a late accepted summary arrives. Any repair here must make the exact checkpoint test and transition one serialized operation, while preserving ambiguous-send custody. Thanks for digging into the handoff.

Maximapple:
Your hold on this is right, and #274 is the narrower version of it.

The objection recorded in the 2026-09-16 review was that this releases ambiguous-send custody on *missing* acceptance signals, which can simply be late. I agree, and #274 does not do that: it acts only on ChatGPT's own rendered transport failure **together with** the page still having no conversation id. That pair is positive evidence rather than absence — it is the invariant the existing comment in `releaseContinuationSourceSendNow` already states, "a chat that still has no id cannot have accepted a message". A failure card that renders for one poll and is gone on the next is ignored, and an id that does arrive outranks the card.

It also bounds the re-offer, which this PR did not: the failure is usually a statement about ChatGPT rather than about one brief, so unbounded re-offering answers a bad afternoon with a fresh chat every couple of minutes for the six hours of the handover TTL.

Happy for this one to be closed in favour of #274 if you prefer a single thread; leaving it to you since the hold was a fair read of what it did.

totec448-spec:
Closing in favor of the already adapted #274 work, as you suggested in your September 17 follow-up. The current continuation owner retains an uncertain dispatched send until exact evidence or explicit cancellation resolves it. The missing-acceptance timeout in this older branch would release that custody, so it is intentionally excluded.

Reviewed head: `f20f006cb16a`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #146, opened 2026-09-09
Make macOS desktop input physical, and say what it hit

Maximapple:
The native half split out of #78, as asked. 19 files, +3,596 / −230, no `docs/`, nothing about browser control.

@hhh2210 — this builds directly on the macOS backend from #28, so your read on the boundary would be worth more than mine.

### What changes

The helper drove input by asking AppKit to synthesise it. Enough to move a pointer and enough to look right in a screenshot, not enough for an application to believe: a drag arrived as a jump, a scroll as a single notch, and a click could land at a point the window had since moved away from.

- **Drag and scroll are spelled out.** An explicit step budget and pointer path instead of one synthesized move; scroll carries its own fraction. The pointer image is read back, so a reply can say what was actually under it.
- **Every physical mutation revalidates its window-bound frame.** A window that moved between observation and action refuses rather than hitting whatever is now at those coordinates.
- **Targeting is fail-closed.** Workspace, WindowServer and AX must agree before physical input is delivered. Explicit-UI and captured-pixel target identities stay apart instead of being merged. System shortcuts route globally; a disabled semantic control is refused. Screen frames bind to the exact active-display topology rather than assuming the primary display.

### Costs, stated

`computer`'s schema grows **6.0k → 6.3k bytes**, and the per-tool ceiling in `test/mcp.test.ts` is raised deliberately rather than the schema trimmed to fit. Spelling a step budget out is precisely what stops the model guessing at one, so the bytes buy something — but the number stays asserted separately from the surface total so the next growth has to argue for itself.

### Also here, because the tests bind it

The disabled-tool refusal now names the **permission** to turn on rather than the internal capability — the difference between a dead end and an instruction. It has its own test.

CI compiles the helper for **both** real macOS architectures, so an arm64-only build can no longer pass while x64 is broken.

### Verified, and its limit

Full suite **3,474 passed**. The macOS assertions here are source-inspection tests, which is how this repo already checks the Swift helper, so they run and pass on any platform — but I should be plain that **I have not run this on a Mac**. The behavioural claims come from the branch it was extracted from; what I verified is that the extraction is faithful and self-consistent.

The four remaining failures are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, which fail identically on unmodified `main` on this Windows machine and pass in CI.

### Scope note

This is one capability area rather than one change, and it is large. If you would rather see it as separate pieces — the pointer/drag work, the fail-closed targeting, the display topology — say which cut you want and I will resubmit along it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,29 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Compile macOS desktop helpers
+        if: runner.os == 'macOS'
+        shell: bash
+        run: |
+          set -euo pipefail
+          node scripts/prepare-macos-desktop-helper.mjs --platform darwin --arch x64
+          node scripts/prepare-macos-desktop-helper.mjs --platform darwin --arch arm64
+
+      # Compiling proves the Swift parses. It says nothing about whether the binary starts,
+      # answers, or reaches its capture path — and the pointer defect that prompted this work
+      # was a runtime property the compiler could not see. This runs it. A runner has no
+      # Screen Recording grant, so a capture is expected to be refused by name; that refusal is
+      # a working code path, while a crash, a hang or a malformed reply is not.
+      - name: Run the macOS desktop helper
+        if: runner.os == 'macOS'
+        shell: bash
+        run: node scripts/probe-macos-helper.mjs arm64
+
+      - name: Limit Vitest workers on Windows
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: Add-Content -Path $env:GITHUB_ENV -Value 'VITEST_MAX_WORKERS=1'
+
       - name: Verify
         run: npm run verify:ci
 
--- a/native/macos-desktop-helper/main.swift
+++ b/native/macos-desktop-helper/main.swift
@@ -214,9 +214,11 @@ private func allWindowRows(includeMinimized: Bool = true) -> [WindowRow] {
         let onScreen = bool(item[kCGWindowIsOnscreen as String])
         let alpha = number(item[kCGWindowAlpha as String])?.doubleValue ?? 1
         guard layer == 0, alpha > 0 else { return nil }
-        let process = string(item[kCGWindowOwnerName as String], default: "Process \(pid)")
-        let title = string(item[kCGWindowName as String]).trimmingCharacters(in: .whitespacesAndNewlines)
-        let displayTitle = title.isEmpty ? "\(process) window" : title
+        let process = boundedAXString(string(item[kCGWindowOwnerName as String], default: "Process \(pid)"))
+        let title = boundedAXString(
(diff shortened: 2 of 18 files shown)


hhh2210:
Thanks for inviting me to check the native boundary. I put the corrections into a follow-up PR against your branch: https://github.com/Maximapple/chat-on-steroids/pull/2 (correction commit `a39c90f` directly on `1f83abc`, plus test-only review follow-up `b0f12e7`).

It separates the macOS pointer z-order check from keyboard-focus recovery, enforces the Windows target-window lease, preserves completion/routes when post-action observation fails, and adds the runtime probe referenced by CI. The existing drag/scroll and focus-recovery behavior stays in scope.

Local evidence: 99 focused tests passed; both macOS architectures compiled; the arm64 capture probe and Electron addon warm-up passed; production build passed. Full verification had one failure, the bundled-`rg` shell-path assertion, reproduced identically on the unmodified head. The full commands, regression evidence and remaining live-input limitations are in the follow-up PR.

GitHub is waiting for maintainer approval before running CI on the follow-up PR. The [equivalent full matrix in my fork](https://github.com/hhh2210/chat-on-steroids/actions/runs/34449921741) completed: macOS and Linux passed; Windows had 3,497 passed and one failure, the unchanged 5,000-folder handoff test exceeding its 90-second timeout. The new Windows target-lease regression passed. I have not changed the unrelated test or represented that Windows job as green.

Both automated review suggestions on the test extractors are addressed and replied to. [Typecheck and the current native regressions passed on both macOS and Windows](https://github.com/hhh2210/chat-on-steroids/actions/runs/34450854916). The proof branches' workflow-only commits are not part of the contribution; the follow-up PR contains the exact scope and remaining live-input limitations.

totec448-spec:
hey @Maximapple, The disabled-permission guidance is already extracted and credited in #186. I read @hhh2210’s follow-up as well; the native input changes need their own macOS runtime acceptance and the current branch checks are not all green. I’m keeping this open rather than claiming Windows source tests prove Mac clicks, drags and focus. Thanks to both of you — love.

totec448-spec:
Reviewed the latest September 19 addition as well as the earlier implementation. The PNG reader and native probe scripts are now included, so the previous missing-script objection is resolved. Your reported ARM64 probe results are useful evidence and are not being dismissed.

Keeping this native refactor open for focused review of the current head against the current Desktop scheduler/focus and input ownership contracts. This Windows review cannot establish native macOS acceptance, and no current-head repository checks were reported by gh pr checks during this pass. The previously incorporated permission guidance remains credited separately.

Reviewed head: `832cc8cb55a8`.

totec448-spec (changes requested):
Reviewed at `832cc8cb55a8`.

This needs a narrow revision against current main and native acceptance for the remaining input boundaries. A moving-window fallback must preserve ambiguity and identity checks. Treating a newline in type_text as physical Return can submit a native composer, so literal multiline insertion needs to remain distinct from an explicit key action. Existing macOS probe results do not cover these negative cases on the complete current head.

PR #157, opened 2026-09-10
Default composer to ChatGPT Native mode to prevent blocking sends on unobserved models

Nir-Bhay:
### Context & Problem
While setting up and using Chat On Steroids on Windows with ChatGPT, I encountered a blocker when trying to send messages from the desktop composer. Every message attempt stalled waiting for model discovery and failed with the toast error:
> *"Model discovery could not confirm your selection. Choose an available model and thinking effort, then send again."*

Checking the extension logs and network traffic revealed that model discovery was returning `picker_unavailable` with 0 models (`models: []`). On accounts using newer ChatGPT subscription configurations or interface variants (such as Go tiers or regional rollout layouts where the DOM model picker is styled differently or absent), the companion extension cannot locate or scrape observed model options from the web page.

Because the desktop composer treated having a verified model catalogue as a mandatory prerequisite for admission, any failure in model discovery made the entire app unable to send messages — even though the local MCP servers, bridge connection, and ChatGPT tab were otherwise functioning properly.

### Root Cause
In `src/renderer/chat.ts`, `sendComposer()` unconditionally waited on model discovery:
```typescript
const modelSettings = confirmedComposerModel() ?? await ensureComposerModel();
if (!modelSettings) {
  toast('Model discovery could not confirm your selection. Choose an available model and thinking effort, then send again.');
  return false;
}
```
When `ensureComposerModel()` returned `null`, sending was rejected outright.

Looking at `extension/chatgpt-dom.js` (around line 1994), however, the companion extension already has native support for sending without an explicit model selection:
```javascript
if (!model && !effort) return true;
```
When `model: null` and `reasoningEffort: null` are passed, the extension bypasses DOM model picker manipulation entirely and injects the message directly into the active ChatGPT conversation. ChatGPT then handles the turn using whatever default model and reasoning settings are already active on the user's account. In addition, the IPC schema in `src/main/session/input.ts` already permits `null` for both `model` and `reasoningEffort`.

### Proposed Fix: ChatGPT Native Mode
Rather than blocking all composer sends on model discovery, this pull request introduces **Native Mode** as the default admission state:

1. **Default to Native Mode (`composerMode = 'native'`)**:
   - The composer defaults to `ChatGPT Default` (both in label and state).
   - In Native Mode, `sendComposer()` bypasses the blocking `ensureComposerModel()` call and submits `{ model: null, reasoningEffort: null }`.
   - Messages are delivered immediately to the active conversation without requiring DOM model picker interaction.

2. **Explicit Selection Mode (`composerMode = 'explicit'`)**:
   - If the user moves the reasoning power slider or chooses an observed model in the dropdown, the composer automatically transitions to Explicit Mode.
   - In Explicit Mode, all existing validation checks, model catalogue confirmations, and error toasts remain active.

3. **User Control**:
   - Added a "Use ChatGPT Default" button (`#composerDefaultBtn`) inside the `#modelMenu` popover so users can return to Native Mode from an explicit selection at any time.
   - Starting a new chat (`selectNewChat()`) cleanly resets to Native Mode, while existing session observations continue to bind properly.

---

## Validation

  - Added `test/composer-native-mode.test.ts` covering 6 deterministic test scenarios:
    1. Native mode with empty catalog (`models: []`) -> sends permitted with `{ model: null, reasoningEffort: null }`.
    2. Native mode with `picker_unavailable` -> sends permitted with `{ model: null, reasoningEffort: null }`.
    3. Explicit mode with `picker_unavailable` -> sends blocked with toast error as expected.
    4. Explicit mode with observed model -> sends with selected `{ model, reasoningEffort }`.
    5. Native <-> Explicit mode switching preserves state, UI labels, and button visibility.
    6. Background catalogue pushes do not reset or overwrite an active Native mode.
  - Updated `test/renderer-chat-models.test.ts` to test explicit catalog reload behavior cleanly.
  - All 36 tests in `test/chat-models.test.ts`, `test/renderer-chat-models.test.ts`, and `test/composer-native-mode.test.ts` pass cleanly.
- [x] `npm run verify` passes (`npm run typecheck` passes with 0 errors).
  - Built the Windows x64 NSIS package with `npm run dist:x64` (`Chat-On-Steroids-Setup-x64.exe`). Tested unpackaged binary in `release/win-unpacked` on Windows; message dispatch operates without stalling on accounts where picker discovery fails.

### DIFF
--- a/src/renderer/chat-models.ts
+++ b/src/renderer/chat-models.ts
@@ -12,6 +12,22 @@ let discovery: Promise<void> | null = null;
 let catalogSubscribed = false;
 type ObservedSelection = { model: string; reasoningEffort?: ReasoningEffort; observedAt: number };
 let composerContext: { scope: string | null; observation: ObservedSelection | null; edited: boolean } | null = null;
+export type ComposerMode = 'native' | 'explicit';
+let composerMode: ComposerMode = 'native';
+
+export function getComposerMode(): ComposerMode {
+  return composerMode;
+}
+
+export function setComposerMode(mode: ComposerMode): void {
+  composerMode = mode;
+  paintStatus();
+}
+
+export function isNativeComposerMode(): boolean {
+  return composerMode === 'native';
+}
+
 const pairs = [['composerModel', 'composerReasoning'], ['workerModel', 'workerReasoning'], ['helperModel', 'helperReasoning']] as const;
 const effortNames: Record<string, string> = { none: 'Instant', minimal: 'Minimal', low: 'Low', medium: 'Medium', high: 'High', xhigh: 'Extra high', max: 'Max', ultra: 'Ultra', pro: 'Pro' } satisfies Record<ReasoningEffort, string>;
 const composerEfforts = ['low', 'medium', 'high', 'xhigh', 'max', 'ultra', 'pro'] as const;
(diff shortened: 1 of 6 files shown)


totec448-spec:
hey @Nir-Bhay, I’m leaving this unmerged because a default implicit native model would change what an explicitly selected model means. It also adds another renderer selection authority, and the escape control is hidden when discovery is unavailable. I’m keeping the account-observed selection contract and fixing discovery at its source. Thanks for proposing a way through the blocked composer.

totec448-spec:
Closing this workaround because defaulting to an unknown/native model bypasses the confirmed account-observed model selection contract. Empty or changed native pickers need an observation fix, not an unconfirmed selection. Existing model-discovery reports remain open for that work.

Reviewed head: `2f48323072aa`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #158, opened 2026-09-10
fix(extension): never close a tab the user pinned

### LINKED ISSUE #150: Pinned ChatGPT tabs can be automatically closed because managed-tab pruning ignores tab.pinned
### App version

2.0.8

Windows 11 x64

### Architecture

x64 (Intel/AMD)

Other or not sure

### What happened

What happened

Chat On Steroids automatically closed a ChatGPT conversation tab even though I had explicitly pinned that tab in Chrome.

The tab was being used as a CoS-managed agent conversation. It disappeared without a normal user close action and did not appear in Chrome’s recently closed tabs.

I expected a user-pinned tab to be protected from every automatic cleanup path. CoS may still mark the underlying conversation as sleeping, finished, retired, duplicated, or superseded, but it should not physically close a tab that the user explicitly pinned.

Source inspection shows that extension/background.js::pruneManagedTabs() can call:

await chrome.tabs.remove(tab.id);

after checking conversation identity, navigation state, unsent work, and the page-side close proof. However, neither the initial tab object nor the fresh current / latest tab objects are checked for pinned === true.

This means a pinned tab is treated exactly like an unpinned tab once its conversation appears in retiredConversations or closableConversations, or when it is considered a duplicate.

This report intentionally does not claim why the observed conversation was classified as closable. That may be a separate lifecycle/identity problem. The narrow confirmed defect here is that the final physical-close boundary ignores Chrome’s pinned state.

Expected behavior:

A tab with pinned === true must never be closed automatically by CoS.
This protection should apply to prime, worker, Goal helper, compacted-source, retired, blocked, and duplicate ChatGPT tabs.
CoS may retain its internal retirement state and close the tab later if the user manually unpins it.
Pinning during the asynchronous close-proof sequence must also prevent closure.

### Steps to reproduce

Steps to reproduce
Open a ChatGPT conversation managed by Chat On Steroids.
Pin its Chrome tab.
Run a workflow that eventually makes the conversation eligible for CoS tab cleanup, such as a worker becoming sleeping/finished or a managed conversation being marked retired/closable.
Leave the extension maintenance process running.
Observe that the pinned tab can be removed automatically.

The exact lifecycle classification that preceded my live occurrence was not captured, but the missing pinned guard is directly visible in the closing path.

Anything else

Relevant code in extension/background.js, function pruneManagedTabs():

const current = await chrome.tabs.get(tab.id);
if (conversationFromUrl(current.url) !== conversationId || current.pendingUrl) continue;

// page-side close proof...

const latest = await chrome.tabs.get(tab.id);
if (
  latest.pendingUrl ||
  conversationFromUrl(latest.url) !== conversationId ||
  !ownsDocument(source) ||
  journalCountForConversation(conversationId) > 0
) continue;

await chrome.tabs.remove(tab.id);

Neither current.pinned nor latest.pinned is checked.

A minimal safe fix would make a pinned tab a hard close veto, including a final race-safe check immediately before removal:

const current = await chrome.tabs.get(tab.id);
if (
  current.pinned ||
  conversationFromUrl(current.url) !== conversationId ||
  current.pendingUrl
) continue;

// page-side close proof...

const latest = await chrome.tabs.get(tab.id);
if (
  latest.pinned ||
  latest.pendingUrl ||
  conversationFromUrl(latest.url) !== conversationId ||
  !ownsDocument(source) ||
  journalCountForConversation(conversationId) > 0
) continue;

await chrome.tabs.remove(tab.id);

Suggested regression tests:

A pinned terminal managed tab is not removed.
The same unpinned terminal tab is still removed.
A tab pinned between the first chrome.tabs.get() and the final close check is not removed.
A pinned duplicate tab is preserved.
Pin protection does not incorrectly make the conversation active or change its internal worker lifecycle state.

### Anything else

```text

```

L4XB:
## Root cause

Fixes #150.

`pruneManagedTabs()` in `extension/background.js` checks conversation identity, navigation state, unsent work and the page-side close proof before `chrome.tabs.remove(tab.id)`, but neither the initial tab object nor the two fresh `chrome.tabs.get` reads are checked for `pinned === true`. A pinned conversation tab was therefore closed exactly like an unpinned one as soon as its conversation showed up in `retiredConversations` / `closableConversations` or counted as a duplicate.

## Smallest behavior change

- `pruneManagedTabs`: a pinned tab is never a close candidate (`continue` before any policy evaluation), and the flag is re-checked on both fresh reads (`current` before the close proof, `latest` right before the close), so pinning during the asynchronous close-proof sequence counts as well. Retirement/closable/duplicate state is untouched; the tab closes once the user unpins it.
- The other automatic closes read a fresh tab object before `chrome.tabs.remove` already; each now also requires `pinned !== true` on that object: the accepted Goal-helper answer close, `retireFailedCommandTab`, the temporary planner retirement, the plugin-refresh tabs and the model-catalog tabs. No permission, identity or recovery logic changes; a pinned tab simply stays open.

## Validation

`test/extension.test.ts`: two new cases run the sliced `pruneManagedTabs` the same way the existing cancelled-helper test does. A retired conversation tab that is already pinned is never messaged or removed and stays in the returned list; a tab that becomes pinned between the close proof and the final read is not removed either. Both fail on `main` (the tab is removed).

```
npx vitest run test/extension.test.ts -t "pinned|cancelled helper"   → 6 passed
npx tsc --noEmit -p tsconfig.json                                    → clean
```

Not exercised against a live Chrome; the change only adds `pinned` checks next to the existing fresh-read guards.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -1252,7 +1252,7 @@ async function retireFailedCommandTab(entry) {
       failedCommand: { id: entry.id, client: entry.client } }, { documentId: source.documentId });
     const latest = await chrome.tabs.get(source.tab);
     if (proof?.safe === true && proof.conversationId === null && proof.navigationEpoch === source.navigationEpoch &&
-        latest && !latest.pendingUrl && latest.url === url && ownsDocument(source)) await chrome.tabs.remove(source.tab);
+        latest && latest.pinned !== true && !latest.pendingUrl && latest.url === url && ownsDocument(source)) await chrome.tabs.remove(source.tab);
   } catch { /* A busy, edited, replaced or unreadable page stays open. */ }
 }
 
@@ -1774,7 +1774,7 @@ async function deliverDesktopInputs(inputs, background, reusableConversations = 
         const proof = await chrome.tabs.sendMessage(tab.id, { type: 'clf-close-temporary-planner', id: input.id, owner: input.owner }, { documentId });
         const current = await chrome.tabs.get(tab.id);
         const successor = replacement ? await chrome.tabs.get(replacement.id) : null;
-        if (proof?.safe === true && ownsDocument(source) && String(current.url || '').includes(marker) &&
+        if (proof?.safe === true && current.pinned !== true && ownsDocument(source) && String(current.url || '').includes(marker) &&
             (input.retire === true || (successor && replacements.some(next => matchesInput(next, successor))))) await chrome.tabs.remove(tab.id);
       } catch { /* only the exact still-owned temporary document may close */ }
       continue;
@@ -1896,7 +1896,8 @@ function inspectRequestedPluginRefresh(publications, background, browserOnly = f
       let timer;
       const proof = await Promise.race([chrome.tabs.sendMessage(tab.id, { type: 'clf-plugin-refresh-state', id }).catch(() => null), new Promise(resolve => { timer = setTimeout(() => resolve(null), 3000); })]).finally(() => clearTimeout(timer));
       if (proof?.safe !== true) return;
-      if (pluginRefreshMarker(await chrome.tabs.get(tab.id).catch(() => null)) === id) await chrome.tabs.remove(tab.id);
+      const latest = await chrome.tabs.get(tab.id).catch(() => null);
+      if (latest?.pinned !== true && pluginRefreshMarker(latest) === id) await chrome.tabs.remove(tab.id);
     }
     if (!requests.length) return;
     const held = tabs.find(tab => requests.some(request => request.id === pluginRefreshMarker(tab)));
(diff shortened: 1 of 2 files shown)


L4XB:
Superseded by `main`, closing.

`main` now guards `tab.pinned` at every close site this PR touched and at six more it did not: the pruning loop, the catalog and plugin refresh paths, the duplicate collapse and both reads around the close proof (17 occurrences in `extension/background.js` against the 9 on this branch). The only thing left in my diff was `pinned !== true` instead of the truthy check, which is not worth a rebase over eight conflicting hunks.

One thing worth separating in #150, since it is still open: the pinned part is fixed, but the newest report there describes a different failure, a freshly created conversation closing while the assistant is still working, on a tab that is neither retired nor duplicate nor pinned. That will not be caught by any `pinned` guard, so #150 probably should not be closed on the strength of this work.

PR #160, opened 2026-09-10
Stop reloading a chat that answers every reload the same way

Maximapple:
A chat whose turn breaks on ChatGPT's side is reloaded every three minutes, indefinitely, with no escalation and no stopping condition — including after the app has itself declared the turn dead.

### Measured

One conversation, one hour of app log, wedged on ChatGPT's own `Connection interrupted. Waiting for the complete answer`:

```
13:49:16  silence-reload        two-minute watchdog fires
13:49:48  transport-failure     the reloaded page reports the same error, 32 s later
13:51:48  silence-reload
13:52:18  transport-failure
…
14:45:35  silence-reload        still going
```

**16 silence reloads, 12 assistant-error reports, one hour.** The decisive detail is at 14:43, where the app recorded

```
turn_end  outcome=stalled  "no visible output and no progress for ten minutes"
```

and reloaded again at 14:45:35. Its own verdict that the turn is dead did not stop the loop. The user's view of the chat shows the pair repeating: `↻ Reloaded chat to recover an unresponsive open turn` followed seconds later by `! Connection interrupted`.

### Why it does not stop

Two triggers reach `queueBrowserRecovery` and only one is bounded. The assistant-error path spends exactly one reload per turn. The silence watchdog has no such budget, deliberately — and the comment above it is right about why:

> Silence asks the chat-level question instead — is this conversation alive at all — and it is the answer to a stuck turn-scoped repair, not something one may mute.

Only `BROWSER_RECOVERY_COOLDOWN_MS` paces it, which is exactly the observed three-minute cadence. The intent is sound and the outcome is a livelock: the remedy is provably ineffective after the first attempt, because the state being recovered from is ChatGPT's rather than the page's.

### The change

**Not a cap.** Silence still runs on every sweep and still supersedes a stuck turn-scoped repair. What stops is repeating one *specific* reload that this chat has already answered the same way three times.

`turnRepairSpent` already models "this remedy is spent for this turn". That is the wrong unit here: a wedged turn never ends, so its budget is never released. The new record is per chat and keyed on the failure **and** the turn, so any different answer clears it — a reload that comes back quiet, a new turn, a different failure. Only the proven-useless remedy is withheld.

It is also said out loud, once, in the chat's own timeline:

> Stopped reloading this chat: 3 reloads each came back with the same failure — "…". The turn is broken on ChatGPT's side, which a reload cannot repair. Continue in a new chat, or send a message here to start a fresh turn.

Sixteen silent retries with no user-visible verdict was the other half of this defect.

### Verified, both directions

| | before | after |
|---|---|---|
| six rounds of the **same** failure | 6 silence reloads, unbounded | 3, plus the assistant-error path's own single per-turn reload |
| four rounds of **different** failures | 4 reloads | 4 reloads, unchanged |

The second row is the one that matters for the objection: the liveness check is not muted. Fail-before output on unmodified `main`:

```
- "assistant-error"
+ "silence"
+ "silence"
```

`npm run typecheck` clean. Full suite 3442 passed; the four failures are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, which fail identically on unmodified `main` on this Windows machine and pass in CI.

### Not included

Escalating to Compact & Resume was the other option considered. It needs ChatGPT to write the brief, and in this state ChatGPT is what is broken — so it cannot be the only path, and it is a larger design decision than this. Recovering an already-wedged chat still means blocking it (which refuses every trigger, since they all converge on `queueBrowserRecovery`) and moving the work by hand.

### DIFF
--- a/src/main/bridge.ts
+++ b/src/main/bridge.ts
@@ -77,6 +77,7 @@ import {
   noteChatOrigin,
   recordAgentMessage,
   recordChatObservations,
+  recordNote,
   recordProgress,
   restoreRecordedConversation,
   setCallAttributionListener,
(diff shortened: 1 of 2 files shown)


totec448-spec:
hey @Maximapple, I’m not taking this algorithm as written. The new counter advances on repeated error observations rather than acknowledged reloads, so replayed events can spend a reload budget without a reload; its map also lacks bounded retirement. A future fix needs to live with the existing recovery owner and count exact receipts. The underlying repeated-reload concern is worth keeping.

totec448-spec:
Closing this older retry-counter proposal. It counts repeated error observations as attempts without proving a browser reload occurred. Current recovery budgets are owned by actual action/receipt state; adding this counter would create a second, disagreeing authority.

Reviewed head: `4c4d31ace7ec`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #161, opened 2026-09-10
Never close a tab the user pinned

### LINKED ISSUE #150: Pinned ChatGPT tabs can be automatically closed because managed-tab pruning ignores tab.pinned

Maximapple:
Closes #150.

### Confirmed

`pinned` did not appear in `extension/background.js` at all:

```
$ grep -c "pinned" extension/background.js
0
```

The reporter's reading is exact: `pruneManagedTabs()` checks conversation identity, navigation state, unsent work and the page-side close proof, then calls `chrome.tabs.remove(tab.id)` — and never asks whether the user pinned the tab.

### Scope

The issue names `pruneManagedTabs` as the narrow confirmed defect and adds:

> I expected a user-pinned tab to be protected from **every** automatic cleanup path.

There were six `tabs.remove` call sites, and every one of them is automatic cleanup — a retired conversation, a spent temporary planner, a plugin-refresh helper, a model-catalog helper, a duplicate, an accepted decision document. None is the user asking for anything. So the rule is stated once, in `closeUnpinnedTab`, and all six go through it:

```
$ grep -c "chrome.tabs.remove" extension/background.js
1
```

It re-reads the tab rather than trusting the caller's object, for the same reason every one of those callers re-reads identity at that point: pinning can happen while the checks above are awaiting.

Deliberately unchanged: the conversation may still be retired, slept, superseded or marked duplicate. This governs the physical close and nothing else. `pruneManagedTabs` now keeps a refused tab in its `remaining` set, so nothing downstream treats it as gone and opens the same conversation beside it.

### Verified

Fails before the change:

```
× never closes a tab the user pinned (pinned: true)
  AssertionError: expected "vi.fn()" to be called +0 times, but got 1 times
```

Passes after, with the unpinned case asserted in the same test so the pruner is not simply disabled.

**One thing worth flagging for review.** Two harnesses (`test/extension.test.ts`, `test/plugin-refresh-workflow.test.ts`) slice these functions out of the worker source and run them in a `vm` context. Both now include `closeUnpinnedTab` in the slice, so they exercise the real close boundary — without that, `plugin-refresh-workflow` failed with a missing reference and the pinned test would have been passing against a stub that could not refuse.

`npm run typecheck` clean. Full suite 3442 passed; the four failures are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, which fail identically on unmodified `main` on this Windows machine and pass in CI.

### Not addressed

Why the reported conversation was classified as closable in the first place. The report is explicit that this may be a separate lifecycle problem, and this PR does not touch that — it only stops the physical close from ignoring a user's claim on the tab.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -1252,7 +1252,7 @@ async function retireFailedCommandTab(entry) {
       failedCommand: { id: entry.id, client: entry.client } }, { documentId: source.documentId });
     const latest = await chrome.tabs.get(source.tab);
     if (proof?.safe === true && proof.conversationId === null && proof.navigationEpoch === source.navigationEpoch &&
-        latest && !latest.pendingUrl && latest.url === url && ownsDocument(source)) await chrome.tabs.remove(source.tab);
+        latest && !latest.pendingUrl && latest.url === url && ownsDocument(source)) await closeUnpinnedTab(source.tab);
   } catch { /* A busy, edited, replaced or unreadable page stays open. */ }
 }
 
@@ -1775,7 +1775,7 @@ async function deliverDesktopInputs(inputs, background, reusableConversations = 
         const current = await chrome.tabs.get(tab.id);
         const successor = replacement ? await chrome.tabs.get(replacement.id) : null;
         if (proof?.safe === true && ownsDocument(source) && String(current.url || '').includes(marker) &&
-            (input.retire === true || (successor && replacements.some(next => matchesInput(next, successor))))) await chrome.tabs.remove(tab.id);
+            (input.retire === true || (successor && replacements.some(next => matchesInput(next, successor))))) await closeUnpinnedTab(tab.id);
       } catch { /* only the exact still-owned temporary document may close */ }
       continue;
     }
@@ -1896,7 +1896,7 @@ function inspectRequestedPluginRefresh(publications, background, browserOnly = f
       let timer;
       const proof = await Promise.race([chrome.tabs.sendMessage(tab.id, { type: 'clf-plugin-refresh-state', id }).catch(() => null), new Promise(resolve => { timer = setTimeout(() => resolve(null), 3000); })]).finally(() => clearTimeout(timer));
       if (proof?.safe !== true) return;
-      if (pluginRefreshMarker(await chrome.tabs.get(tab.id).catch(() => null)) === id) await chrome.tabs.remove(tab.id);
+      if (pluginRefreshMarker(await chrome.tabs.get(tab.id).catch(() => null)) === id) await closeUnpinnedTab(tab.id);
     }
     if (!requests.length) return;
     const held = tabs.find(tab => requests.some(request => request.id === pluginRefreshMarker(tab)));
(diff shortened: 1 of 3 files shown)


totec448-spec:
hey @Maximapple, Pinned tabs are now checked at the automatic close paths, including a pin added while the close check is awaiting its result. Those regressions pass. Your pin-protection proposal is credited; closing this as incorporated without adding another asynchronous gap before close. Love.

PR #164, opened 2026-09-11
Judge a replacement chat's checkpoints on its lease, not on its route

Maximapple:
**This is the one that produces the common symptom: a replacement chat opens, and nothing is ever typed into it.**

### Measured, from the replacement tab itself

Shipped extension 2.0.8, 2026-09-10:

```json
{"at":"04:39:55.218","type":"redeem","ok":true,"hasCommand":true,"textLen":48975,"url":"/?clf=42c8b63e3001a081"}
{"at":"04:39:58.262","type":"compact","flags":["destinationAttempt"],"ok":false,"error":"stale_document","url":"/?clf=42c8b63e3001a081"}
```

The brief was delivered in full and thrown away three seconds later. Four consecutive handoffs failed exactly this way.

### Mechanism

`content.js` asks the app for permission before the irreversible send, and reads any non-ok reply as a denied permit:

```js
const permit = await ask({ type: 'compact', token, destinationAttempt: true });
if (!permit || permit.ok !== true || !permit.data || permit.data.allowed !== true) {
  CLF_DOM.clearPromptExact(boot.text);
  return;                               // silent: no ack, no log, empty composer
}
```

The service worker never forwards it. Its route guard:

```js
if (!ownsDocument(source) || !tab || tab.pendingUrl || tab.status === 'loading' ||
    !isChatGptUrl(tab.url) || conversationFromUrl(tab.url) !== cleanConversationId(message.conversationId))
  return { ok: false, error: 'stale_document' };
```

That is written for checkpoints that *name* a chat: the tab has to prove it settled on that exact route. The two destination checkpoints name none — they come from a page opened at `/?clf=<id>`, before ChatGPT has created anything — and a ChatGPT opened seconds ago is still `loading`. So the normal case is refused.

Nothing downstream can see it, because the refusal is the worker's own and no request reaches the app. `fail()` would travel the same call, so no ack escapes either. The app waits out its deadline and reports `the chat this app opened did not report back in time`, which names the timeout rather than the cause.

That timing is also why it is intermittent rather than total: whether the tab has finished loading when the permit is asked for depends on the machine and the network.

### The change

The guard splits on what the message claims. A checkpoint naming a chat still has to prove the tab settled on that route — **unchanged**. One naming none is judged on the document lease `ownsDocument` has already established, and refused only if the tab turns out to be showing some other conversation.

21 lines in `extension/background.js`.

### Verified

Fails before the change:

```
× forwards a replacement chat permit while the tab is still loading
× forwards a replacement chat permit while the tab is still pending
  AssertionError: expected Object{ ok: false, …(1) } to not match object { error: 'stale_document' }
```

Passes after, plus `forwards the destination checkpoints a replacement chat sends, which name no conversation`.

**Worth flagging:** every prior `/compact` test passes `conversationId` alongside the token, because every one of them is the *source* chat. The destination's shape — token and flag, no conversation — was untested, and it is the shape the whole handoff depends on.

Full suite 3442 passed. Four of the five failures are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, identical on unmodified `main` here; the fifth, `codex-runtime-parity`, passes 17/17 in isolation and is a load flake on this machine — none of them touches the extension.

### Provenance

The measurement and the fix are from a field investigation on macOS; the report lists this as **verified in production**. I verified the fail-before/pass-after on `main` on Windows before submitting.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -2849,8 +2849,27 @@ const HANDLERS = {
     // ChatGPT assigns /c/B through an SPA transition. Read Chrome's current tab and
     // retain the exact document/epoch lease across that await before accepting its route.
     const tab = await chrome.tabs.get(source.tab).catch(() => null);
-    if (!ownsDocument(source) || !tab || tab.pendingUrl || tab.status === 'loading' ||
-        !isChatGptUrl(tab.url) || conversationFromUrl(tab.url) !== cleanConversationId(message.conversationId))
+    if (!ownsDocument(source) || !tab || !isChatGptUrl(tab.url || tab.pendingUrl))
+      return { ok: false, error: 'stale_document' };
+    const named = cleanConversationId(message.conversationId);
+    // A checkpoint that names a chat has a route to verify, and an unsettled tab cannot prove
+    // it: wait for the navigation rather than accept a message about a chat this tab may be
+    // leaving. A checkpoint that names none is the opposite case and has to be judged
+    // differently — the two destination checkpoints come from a replacement chat that ChatGPT
+    // has not created yet, so there is no id to compare and no navigation to lose.
+    //
+    // Refusing those on `loading` cost a whole handoff every time the fresh tab was slower than
+    // the page asking for its permit. Measured on 2026-09-10: the tab redeemed its brief at
+    // 04:39:55.218 and asked for `destinationAttempt` at 04:39:58.262, three seconds into a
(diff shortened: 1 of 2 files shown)


totec448-spec:
hey @Maximapple, The loading-destination permit fix is already adapted on main, with an extra check for a conflicting pending navigation. I reran the replacement-chat tests. Closing this as incorporated, with your original contribution credited. Thanks — love.

PR #165, opened 2026-09-11
Release a handoff's claim with the command that held it

Maximapple:
An automatic compaction opens tab after tab and never types into any of them, for the ticket's whole six-hour life.

### Mechanism

Redeeming a resume is what *claims* its brief. `claimContinuationNow(token, claimant)` records the redeeming **command id** as `claimedBy`, so a second tab on the same marker cannot be handed the same handoff twice:

```ts
// src/main/session/continuation.ts
if (entry.claimedBy !== null && entry.claimedBy !== claimant) return null;
```

An *automatic* ticket then deliberately outlives a failed attempt. `drop()` retires the command but keeps the ticket, precisely so a later pickup can try again:

```ts
if (automaticResume) {
  commands = commands.filter((entry) => entry !== command);
  logWarn(`bridge: released ${specKey(command.spec)} browser attempt without closing its ticket — ${why}`);
  …
}
```

The claim was **not** retired with it. Every later pickup is a fresh command id, which can therefore never equal the `claimedBy` the dead command left behind — so `claimContinuationNow` refuses for the rest of the TTL, the command is handed out with an empty brief, and the page stops without typing, without an ack and without a log line.

### Field evidence

`state/continuations.json`, hours after the first tab was closed:

```json
{
  "state": "claimed",
  "claimedBy": "253b3fcea4dce828",
  "sourceSend":      { "state": "sent", "messageId": "…" },
  "destinationSend": { "state": "not-attempted", "conversationId": null, "messageId": null }
}
```

`253b3fcea4dce828` was the command from pickup 1; the live command at that moment was `1cf2cea330d67876`. `destinationSend` never left `not-attempted`, so nothing was ever submitted under that claim.

The app log, heartbeats filtered — four attempts, no acks, no failures:

```
20:24:51  captured the compaction brief for <session>; opening the replacement chat
20:39:52  released resume:<session> browser attempt without closing its ticket — the chat this app opened did not report back in time
20:40:04  compaction ticket … opening pickup 1 of 3
20:55:06  released … did not report back in time
20:55:35  compaction ticket … opening pickup 2 of 3
```

### The change

`releaseContinuationDestinationSendNow` is already exactly this transition, and its own comment states the requirement:

> The claim goes with the dispatch. It named the one command whose page was to send the brief; that page has sent nothing and its command is retired, so the next command — a fresh id — must be able to claim, or the released brief could never be offered again.

It was only ever reached from the `/compact destinationLost` route — which a page sends **after surviving its send**. A page that dies before one never sends it. `drop()` now takes the same transition.

Gated on `destinationSend` still being `not-attempted`, so it can only release a brief that provably was not submitted. A dispatched one is left exactly as before, because releasing that is the double send the function already refuses for `sent`.

### Verified

Fails before the change on `main`:

```
× lets the next pickup claim a brief whose first chat died before typing anything
  AssertionError: redeem 4ab768436548d43b failed: expected 409 to be 200
```

The test drives the real bridge: an automatic ticket, a page that redeems and then never reports, the command's deadline, and a second pickup that must still carry the same brief.

`npm run typecheck` clean. Full suite 3442 passed. Of the six failures, four are `renderer-usage` (locale-dependent number formatting) and one `mcp` shell test, which fail identically on unmodified `main` on this Windows machine; the other two (`codex-runtime-parity`, `session`) pass 152/152 in isolation and are load flakes here. None touches this code.

### Related

This is one of several defects from the same investigation. #164 covers the one that produces the visible symptom most often — a replacement chat that opens and stays empty — and is independent of this.

### DIFF
--- a/src/main/bridge.ts
+++ b/src/main/bridge.ts
@@ -7080,7 +7080,8 @@ function describe(command: Command, client: string | null, claimedSummary?: stri
 
 function drop(command: Command, why: string): boolean {
   if (!commands.includes(command)) return false;
-  const automaticEntry = command.spec.type === 'resume' ? continuationByToken(command.spec.token) : null;
+  const resumeToken = command.spec.type === 'resume' ? command.spec.token : null;
+  const automaticEntry = resumeToken ? continuationByToken(resumeToken) : null;
   const automaticResume =
     automaticEntry?.automatic === true && automaticEntry.state !== 'committing' && automaticEntry.state !== 'committed';
   if (automaticResume) {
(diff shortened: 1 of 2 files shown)


totec448-spec:
hey @Maximapple, This is already adapted on main: command retirement releases the original claim inside the serialized transition, with the claimant checked again under the continuation lock. The existing regression passes. Closing the incorporated branch; your credit stays with the fix. Love.

PR #124, opened 2026-09-09
Add minimal durable external control and exact-turn cancellation

ventianima-lab:
A trusted local controller needs to submit an instruction and retrieve its exact app-recorded result without taking ownership of Chat On Steroids browser, session, retry, or recovery policy.

This PR adds a thin authenticated loopback adapter:

- 127.0.0.1 discovery and bearer-token transport
- read-only connection, model, project, and session discovery
- submit, status/result, and exact-turn cancel routes
- request IDs reuse the stock durable input outbox ID
- results come only from the stock session recorder
- automatic Compact & Resume lineage follows exact source-handoff and destination-bootstrap message IDs
- adapter-only continuation metadata is retained for the same 24-hour idempotency window
- a read-only continuation observer records stock transitions without changing their outcome

The adapter does not reload browser pages, replay prompts, create replacement sessions, or add an alternate Pro route. Chat On Steroids still owns delivery, browser wake, session/turn lifetime, retries, Compact & Resume, and recovery.

Exactness gates:

- a later user message ends ownership of the original request
- internal handoff summaries are never returned as task results
- pending continuation cancellation stops the exact owned turn before durably aborting the ticket
- a resumed final is completed only with destination message-scoped model/reasoning evidence
- stock Compact & Resume currently lacks that evidence in some new chats; those requests fail closed and upstream issue #155 tracks the lifecycle fix

Addresses #82.
Addresses #136.
Addresses #151.
Addresses #153.

Validation:

- `npm run typecheck`
- 82 focused control/continuation tests passed
- production build passed
- live Windows install `890bb98`
- active unpacked extension: 14/14 files match upstream 2.0.8
- Core, Desktop, and Plugins live after hidden restart
- completed Coin request retained its exact ID, verified `5.6/pro`, and full result
- two undelivered timeout requests remained terminal with `deliveredAt=null`
- automatic-continuation request no longer reports false cancellation; missing resumed model evidence is reported explicitly

PR #120, opened 2026-09-09
External controller API for persistent ChatGPT workers

reed6868:
Implements the smallest useful scope of #82: a separate authenticated local controller surface for persistent ChatGPT workers, without requiring or synthesizing a ChatGPT Prime conversation.

## Scope

- external ownership keyed by `(controllerId, workerKey)`;
- durable caller-supplied `operationId` for mutating controller operations;
- exact replay returns the accepted operation; conflicting reuse is rejected;
- stage -> durable write -> browser publication ordering;
- explicit `ambiguous` state for delivery uncertainty;
- loopback-only `127.0.0.1` API with an independent generated bearer capability and POSIX `0600` capability file;
- browser-origin requests are rejected;
- existing browser worker bootstrap/revival transport is reused;
- existing Prime-owned `agents` behavior remains preserved behind the ownership facade;
- external ownership is restored before bridge command recovery;
- API/contract documentation in `docs/external-controller.md`.

## HTTP surface

- `POST /v1/workers/ensure`
- `POST /v1/workers/inspect`
- `POST /v1/workers/send`
- `POST /v1/deliveries/inspect`

This intentionally does not add scheduling, workflow graphs, review/approval semantics, artifact management, or provider routing.

## TDD history

Tests were committed before the corresponding implementation milestones:

- `da1f954` — external controller identity/idempotency contract;
- `b0959d6` — loopback/authentication contract;
- `329f102` — bridge reuse contract;
- `40271ab` — durable publication boundary contract.

## Fork validation

The feature branch was fully validated in the fork before this upstream PR was opened.

Fork CI run: https://github.com/reed6868/chat-on-steroids/actions/runs/34299655114
Head validated: `194a4d02b9f283c0b47143a27adf57ea745f74a6`

All repository CI jobs completed successfully:

- Verify Linux x64 — success
- Verify macOS arm64 — success
- Verify Windows x64 — success

Each platform completed `npm run verify:ci` and the published-plugin live tests successfully.

This replaces the earlier draft #118, which was intentionally closed so fork-side full CI could be completed before a formal upstream submission.

Refs #82

Issue #82, opened 2026-09-06
Feature: external controller API for persistent worker conversations


winnal:
### Problem or workflow

# Feature request: External controller API for persistent worker conversations

## Summary

Chat On Steroids already has a useful worker runtime behind the `agents` tool: it can create ChatGPT worker conversations, keep them around, wake them later, deliver additional work, observe lifecycle state, and preserve them across app restarts.

Today, however, that worker runtime is tightly coupled to the assumption that a ChatGPT conversation is the controlling "prime". This makes it difficult for other local tools, agent frameworks, IDEs, workflow engines, or orchestration systems to reuse CoS as a persistent ChatGPT worker runtime.

It would be valuable to expose a small, authenticated **external controller API** alongside the existing Prime-owned `agents` API.

The goal is not to change the current `agents` behavior. The existing Prime → worker model should continue to work exactly as it does now.

Instead, CoS would expose the underlying worker/browser lifecycle as a generic local capability that another trusted application can use.

## Motivation

There are two distinct use cases for CoS workers:

### 1. ChatGPT-controlled swarm

The existing model:

```text
ChatGPT Prime
    ├── worker-1
    ├── worker-2
    └── worker-3
````

The Prime owns the worker family and controls it through the `agents` MCP tool.

This is useful and should remain unchanged.

### 2. Externally-controlled persistent workers

A local application may instead want CoS to provide the ChatGPT worker runtime while the application itself owns higher-level worker identity, lifecycle, scheduling, communication, or workflow semantics.

For example:

```text
Local controller
      │
      ▼
CoS worker runtime
      │
      ▼
Persistent ChatGPT conversation
```

In this mode, the external application should not need to pretend to be a ChatGPT Prime conversation or manufacture a fake `primeConversationId`.

The controller should be able to address a worker directly using a stable externally supplied key.

This would make CoS reusable by many kinds of local agent systems without forcing those systems to adopt CoS's Prime/swarm topology.

---

## Proposed architecture

Keep the existing Prime-owned broker intact and add a second ownership mode:

```text
CoS
├── Prime-owned worker families
│   ├── Prime conversation A
│   │   ├── worker-1
│   │   └── worker-2
│   └── Prime conversation B
│       └── worker-1
│
└── Externally-controlled workers
    ├── controller A / worker X
    ├── controller A / worker Y
    └── controller B / worker Z
```

Both modes can share the same lower-level browser/ChatGPT machinery.

[...]

PR #166, opened 2026-09-11
Add thin authenticated controller ingress

ventianima-lab:
Codex currently has no narrow, durable way to submit work to Chat On Steroids and retrieve the exact recorded reply. Integrations therefore tend to duplicate browser/session lifecycle logic and become fragile across app updates.

This adds a loopback-only authenticated controller ingress that delegates command delivery to the existing durable desktop input queue and joins results through the existing recorder using the request UUID. The API is limited to capabilities, submit, status/result, and pre-send cancellation. It does not open or manage browser sessions, discover models, recover turns, or coordinate agents.

Security and identity properties:
- binds only to 127.0.0.1 on an ephemeral port
- uses a 256-bit bearer token stored beside a versioned discovery descriptor
- rejects browser-origin requests and incompatible protocol versions
- makes the caller-provided UUID the durable idempotency key
- rejects same-ID/different-payload submissions
- returns a result only after a confirmed recorded user row and a final assistant row in that exact request boundary
- recovers a lost delivery ACK only when one recorder-owned user row matches the exact authorized input in a bounded window
- when the target session is already known, searches only that exact session; when it is unknown, duplicate candidates fail closed

The late-recorder recovery addresses #168 without resubmitting an ambiguously accepted prompt.

Validation:
- `npm test -- --run test/control.test.ts` (11 passed)
- `npm test -- --run test/control.test.ts test/input-delivery-integration.test.ts test/session-input.test.ts` (172 passed before the known-session extension)
- `npm run typecheck`
- full Vitest suite with two workers (3,449 passed, 29 skipped) plus `test/mcp-shutdown.test.ts` (2 passed)
- `npm run build`

The ordinary `npm run verify` run hit an existing Windows shell test timeout under full parallel load; that exact test passed alone in 2.18s and the complete suite passed with two workers.

### DIFF
--- a/src/main/control.ts
+++ b/src/main/control.ts
@@ -108,7 +108,12 @@ export async function controlRequestView(deps: ControlDependencies, entry: Input
   };
   if (entry.state !== 'sent' || !sessionId) return base;
 
-  const events = await deps.events(sessionId);
+  // Request/result ownership follows the recorder's durable sequence, never provider clocks.
+  // A freshly opened ChatGPT page can report its final DOM snapshot with an earlier provider
+  // timestamp than the delivery acknowledgement even though the recorder assigned it the next
+  // sequence. Presentation chronology is useful in the UI, but would strand that exact answer
+  // before its owning user message here.
+  const events = (await deps.events(sessionId)).sort((left, right) => left.seq - right.seq);
   const authored = events.findIndex(event => event.kind === 'user_message' && event.inputId === entry.id &&
     event.inputDelivery === 'confirmed');
   if (authored < 0) return base;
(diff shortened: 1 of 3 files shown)


totec448-spec:
hey @ventianima-lab, The thin ingress is the better direction, but I’m leaving it unmerged. Matching input across sessions by time and backtick-stripped text is not a send receipt, and an unowned final cannot become this request’s result just because it falls in a sequence range. A narrower revision should consume the stock outbox’s exact accepted-message/turn relationship. Thanks for working through this.

totec448-spec:
Keeping this as the scoped external-controller proposal for #82; the older overlapping #120 and #124 branches are now closed. The remaining review boundary is exact response ownership: time/text matching or an inclusive journal sequence range must not let another response become this request’s final result. Cancellation also needs to distinguish a queued Stop from native Stop receipt. Please retain those negative cases in the controller integration tests.

Reviewed head: `e443ace2fa57`.

totec448-spec (changes requested):
Reviewed at `e443ace2fa57`.

External requests need exact outbox and response ownership. Time windows, normalized/backtick-stripped text and journal spans can associate a final with another overlapping request. The always-started HTTP surface also needs an explicit configuration and shutdown contract. Keep this open for a narrower controller using existing input/receipt ownership; the current matching is not sufficient to merge.

PR #175, opened 2026-09-11
Fix reload final leaving prime turn active

nofihq:
## Summary
- recover an already-open prime turn when reload reports its current final answer under a distinct message identity
- carry the bounded set of server request IDs for multi-request turns instead of discarding identity
- require exact request-set equality plus proof that the set belongs to exactly one Fiber page turn, so Retry/regenerate request reuse cannot close newer work
- retain the newer-user, running-tool, explicit-end, and durable work-order fences before closing the turn

## Live reproduction
A fresh GPT-6 Pro chat called the session tool, was reloaded for delayed request attribution, then rendered a final answer. The final answer had a different provider/message identity from the pre-reload commentary. It was recorded with no turn ID, leaving the original turn active indefinitely and blocking queued resume input.

## Verification
- focused content/recorder regression run: 557 passed
- focused bridge boundary run: 1 passed, 315 skipped
- full npm run verify: 3,818 passed, 131 skipped
- typecheck, public-history privacy, third-party notices, native-source packaging, and Electron load checks passed

### DIFF
--- a/extension/content.js
+++ b/extension/content.js
@@ -3197,11 +3197,18 @@
    * position or "nearest turn" guess takes part: the answer is a single turn id or nothing.
    */
   function settledTurnOwner(turn) {
-    const requests = new Set();
-    for (const call of (turn && turn.calls) || []) if (call && call.requestId) requests.add(call.requestId);
-    if (requests.size !== 1) return null;
-    const requestId = requests.values().next().value;
+    const requests = requestIdsForTurn(turn);
+    if (requests.length !== 1) return null;
+    const requestId = requests[0];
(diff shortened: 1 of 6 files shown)


nofihq:
@codex review

nofihq:
@codex review

nofihq:
@codex review

nofihq:
Withdrawing this approach. Review exposed three unresolved identity boundaries (Plugins requests, durable late-repair ownership, and unreadable live Fiber descriptors), and upstream main has since advanced with request-origin/send-evidence changes. The patch will not be installed or merged.

PR #179, opened 2026-09-11
Read a nested authored block once, not once per enclosing block

Maximapple:
`messageText()` collects a user message's `.whitespace-pre-wrap` blocks with `querySelectorAll`, which also returns a match nested inside an earlier match, and `text()` reads a whole subtree. An inner block is therefore read twice: once as part of its container and once on its own account.

The recorded message then carries that passage a second time, so every reader that compares authored text against what was submitted sees a message longer than the one it sent.

Measured on a resumed handoff: the rendered bubble came back **9079 characters longer** than the 60425 that were sent, and the exact-text comparison refused to recognise it. An established chat, where nothing nests, compared cleanly in the same data set. Reproduced in the harness: a nested 9000-character block on a 59000-character message yields exactly `diff=+9000`.

`node` itself never counts as a container — the query cannot return it, and treating it as one would empty this preferred path for any message whose own element carries the class.

Test in `test/content-script.test.ts`, red before the change (`…the quoted passage` appended a second time), green after. 828 tests pass across `content-script`, `chatgpt-dom-input`, `extension`, `fiber` and `plugin-refresh-dom`.

### DIFF
--- a/extension/chatgpt-dom.js
+++ b/extension/chatgpt-dom.js
@@ -174,7 +174,19 @@ var CLF_DOM = (() => {
     return safe(() => {
       if (!node) return '';
       if (role === 'user') {
+        // Only blocks that nothing else here already contains. `querySelectorAll` also
+        // returns a match nested inside an earlier match, and `text()` reads a whole
+        // subtree, so an inner block was read twice: once as part of its container and
+        // once on its own. The recorded message then carried that passage twice, and every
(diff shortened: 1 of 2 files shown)


totec448-spec:
hey @Maximapple, Your nested-text fix and regression are already incorporated on main. I checked the actual public source and reran the test, so I’m closing the duplicate branch rather than merging it twice. Your original PR and authorship remain credited. Thanks for catching this — love.

Issue #210, opened 2026-09-13
Agents: expose an evidence-based read-only health projection


Inmerson:
## Problem / workflow

The worker runtime already owns several strong pieces of liveness and lifecycle evidence, but callers have to interpret them independently:

- broker lifecycle (`invited`, `active`, `detached`, `waking`, `sleeping`, terminal states)
- exact conversation ownership / identity
- recorder evidence such as `generating` and `activeTurnId`
- `runningToolCalls(conversationId)` for work still inside MCP dispatch
- browser/tab presence and detach evidence
- existing bounded waits / delivery state where applicable

This makes it difficult for UI/status surfaces and future recovery logic to answer a narrower question consistently: **what does the current evidence prove about this worker right now?**

A fork prototype has been useful here, but I do not think the upstream change should import its Control Center or automatic recovery machinery. The smallest useful upstream slice is a pure, read-only projection over evidence upstream already owns.

## Proposed first slice

Add a small side-effect-free classifier, conceptually:

```ts
type AgentActivity =
  | 'starting'
  | 'working'
  | 'tool_call'
  | 'waiting'
  | 'sleeping'
  | 'done';

type AgentHealth =
  | 'healthy'
  | 'degraded'
  | 'stalled'
  | 'blocked'
  | 'unknown';

type AgentHealthRecommendedAction =
  | 'none'
  | 'observe'
  | 'retry_delivery'
  | 'user_attention';

interface AgentHealthEvidence {
  identity: 'exact' | 'missing' | 'conflicting';
  browser: 'present' | 'missing' | 'unknown';
  runningToolCalls: number;
  generating: boolean | null;
  activeTurn: boolean | null;
  wait: null | {
    kind: string;
    deadlineAt: number | null;
    expired: boolean | null;
    exempt: boolean;
  };
  blocker: null | {
    kind: string;
    summary: string;
  };
}

interface AgentHealthSnapshot {
  agentId: string;
  conversationId: string | null;
  activity: AgentActivity;
  health: AgentHealth;
  recommendedAction: AgentHealthRecommendedAction;
  reason: string;
  observedAt: number;
  evidence: AgentHealthEvidence;
}
```

The exact names can differ. The important part is that the evaluator is **pure and read-only**: it receives already-authoritative evidence and returns a bounded explanation. It must not reload tabs, wake workers, retry delivery, mutate broker state, or introduce a second timer/state machine.

[...]

PR #221, opened 2026-09-14
feat(agent-health): add read-only evidence projection

Inmerson:
Implements the first read-only slice proposed in #210.

## Problem

The broker, recorder and MCP call-context already own strong pieces of worker liveness/lifecycle evidence, but consumers otherwise have to interpret those facts independently. That invites inconsistent or overconfident classifications, especially around detached browser state, exact identity, active tool work and finite waits.

## Smallest behavior change

Adds a pure deterministic `evaluateAgentHealth` projection in `src/shared/agent-health.ts` plus focused regression tests.

- explicit durable blocker evidence takes precedence;
- missing/conflicting exact conversation identity fails closed to `unknown`;
- exact running tool work, generation or an active turn remains healthy active work even when weaker browser/timing evidence looks stale;
- sleeping and terminal broker states are healthy non-running lifecycle states;
- only an owner-provided finite wait deadline may produce `stalled`;
- exempt/unbounded waits do not manufacture stalls;
- detached/browser-missing evidence may degrade confidence but does not claim a server-side turn ended;
- malformed timing evidence degrades instead of fabricating precision;
- human-readable reason/blocker evidence is bounded.

This PR intentionally does **not** add a renderer/API adapter, state owner, timer, watchdog, polling loop, tab reload, worker wake or automatic retry. Evidence ownership stays with the existing subsystems; this file only classifies supplied facts.

## TDD / validation

RED CI run `34817849917` on the contract skeleton:
- 4,248 existing tests passed;
- all 10 new `agent-health` tests failed only because `evaluateAgentHealth` still threw `agent health projection not implemented`.

GREEN CI run `34818265186`:
- macOS arm64: Verify + published-plugin live smoke ✅
- Linux x64: Verify + published-plugin live smoke ✅
- Windows x64: Verify + published-plugin live smoke ✅
- Linux exact run: 4,258 passed / 134 skipped in the primary suite; isolated shutdown 2/2; `agent-health` 10/10; plugin live smoke 5/5
- privacy verification, notices and TypeScript typecheck all passed.

Refs #210


Inmerson:
@totec448-spec this is ready for review. It implements only the first read-only slice from #210; all CI jobs are green across macOS, Linux, and Windows.

k-spit:
that one is massive

totec448-spec:
The classifier is a useful direction for #210, but this head adds standalone health logic without a production consumer in the worker UI. Keeping it open for that focused read-only integration and the corresponding visible-state tests. It should continue to project existing broker evidence rather than become another recovery or lifecycle owner.

Reviewed head: `140ea4da2ac6`.

totec448-spec (commented):
Reviewed at `140ea4da2ac6`.

The read-only health projection is a reasonable shape, and this review does not treat its intentional lack of mutations as a defect. Keeping it open because the new classifier currently has no production consumer. A focused connection to the existing worker pane would make the information usable while preserving the broker as the sole lifecycle owner.

Issue #218, opened 2026-09-14
Compaction: "the replacement chat belongs to another local session"


ferrarinobrakes:
### App version

21.0

### Operating system and version

24h2 windows 11

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT workspace / custom MCP access

Pro (custom MCP read/fetch only)

### What happened

There are times when CoS reaches the token limit warranting compaction, it sends the prompt, receives handoff back, opens a new tab and sends it back but instead of the same conversation in-app continuing, a new one is created and the error is "the replacement chat belongs to another local session" which makes no sense whatsoever. I don't fiddle with it, close tabs, open tabs for it. It makes it lose all the workers it remembered and wasn't a thing before 2.1.0 really.

### Anything else

```text

```

Terryli246:
Reopening with a reproducible recurrence on the current latest release **v2.1.12** on macOS.

This is the same failure signature as the original report. In one working session today, Compact & Resume failed twice in succession, and the same collision also occurred earlier in a separate session.

### Current reproduction / timeline

Config:
- `sessions.advisoryTokens = 400000`
- `sessions.limitTokens = 533333`
- `compaction.auto = true`
- `compaction.autoTokens = 400000`

First automatic attempt:
- source was working at **402,481 context tokens**
- continuation `H_9gfqaV...` was opened
- handoff was prepared and captured successfully (**35,961 chars**)
- COS opened a replacement ChatGPT conversation
- recorder immediately created a new local session for that replacement conversation
- it was then marked `origin.kind=resume` (`session ... named for the resume chat this app opened`)
- the continuation commit failed with:
  - `the replacement chat already belongs to another local session`
- resume completed as `terminal-failure`

Because the source session never actually moved off the original conversation, it remained above the auto-compaction threshold. About 6 minutes later COS filed another automatic ticket against the same source at **413,496 context tokens**:
- continuation `KVw7xtvW...`
- handoff captured successfully (**22,786 chars**)
- replacement conversation opened
- a fresh local session was again created/named for the resume chat
- commit failed with the same `the replacement chat already belongs to another local session`

There was also an earlier occurrence in the same app log today:
- continuation `lhsQUZMT...`
- handoff captured successfully (**28,755 chars**)
- replacement session created and named as resume
- same collision error and terminal failure

### Source-level evidence

[...]

Issue #212, opened 2026-09-13
AutoCompact Issues


Lobsterpaws-77:
### App version

.9

### Operating system and version

osx 

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT workspace / custom MCP access

Other or not sure

### What happened

Auto Compact was intermittently failing on a very large ChatGPT Project conversation.

The compaction step itself worked: COS generated and saved the handoff successfully. The failure happened afterward, when COS tried to open the replacement chat inside the same ChatGPT Project.

The failure message was:

"ChatGPT could not open the source Project through its native link; nothing was sent"

A duplicate/source tab would usually be left open temporarily and would sometimes close itself later. No handoff was sent to a new conversation because COS correctly failed closed.

This became especially reproducible on a large Project conversation at roughly 600k context tokens, with handoffs around 45-47 KB.

Expected behavior:
After generating the handoff, Auto Compact should enter the same Project, create/open a fresh Project chat, send the handoff there, and commit the continuation without requiring user intervention.

### Steps to reproduce

1. Open a large ChatGPT conversation that belongs to a Project and enable Auto Compact.

2. Let the conversation grow enough for Auto Compact to trigger, or manually run Compact & Resume on the same large Project conversation.

3. COS successfully generates the handoff, then opens a marked copy of the source conversation in preparation for entering the Project.

4. On affected runs, the replacement process fails roughly 10-20 seconds later with:
   "ChatGPT could not open the source Project through its native link; nothing was sent"

5. The original/source Project conversation remains intact. A temporary duplicate tab may remain open for a while and may later close automatically.

The problem was intermittent: the exact same Project and workflow could occasionally succeed, which made it appear timing-dependent rather than a permanently invalid Project URL or handoff.

### Anything else

```text
We investigated this fairly extensively and were able to isolate and locally fix the problem.

The handoff size was not the immediate failure. COS was failing before it ever attempted to insert or send the handoff into the destination chat.

The actual problem was a race during ChatGPT Project navigation.

COS opened a duplicate of the source conversation and waited for the native Project-home link to become usable. On a large conversation, ChatGPT's UI can take a significant amount of time to hydrate.

There were two related timing problems:

1. The initial source-page hydration window was only 12 seconds. On this large conversation, that was sometimes not enough for the composer and Project navigation UI to become fully ready.

[...]

Issue #167, opened 2026-09-11
Compact & Resume replacement can stay claimed before destinationAttempt until tab reload


Bemirror99:
## Summary

A Compact & Resume replacement tab can successfully redeem/claim its resume command, then make no further progress before the first destination checkpoint. The new ChatGPT page is fully loaded, the composer is empty and usable, and the COS content script is present, but the continuation remains indefinitely at:

```text
state = claimed
destinationSend = not-attempted
to = null
```

Reloading that exact replacement tab lets a new content-script document safely take over the same still-unattempted command, after which the bootstrap is sent and the continuation commits within a few seconds.

This reproduced twice in a row during one Compact & Resume recovery attempt.

## Environment

- Windows 11
- Chrome 153
- Chat On Steroids desktop app: installed/running v2.0.8
- unpacked extension files updated to current upstream `main` at `53d5aaa` and reloaded before the test
- current upstream extension therefore includes #141 / `destinationLost` forwarding

## Observed sequence

A fresh manual Compact & Resume was started for an existing long-running session.

1. Source handoff completed normally.
   - handoff prepared: 38,933 characters
   - source send checkpoint: `sent`
2. COS opened the replacement ChatGPT tab.
3. The replacement page redeemed the resume command and the durable command became leased/claimed.
4. The continuation stayed at `destinationSend = not-attempted`.
5. The replacement URL retained its exact `?clf=<command-id>#clf=<command-id>` marker.
6. The page itself was healthy:
   - ChatGPT new-chat UI loaded
   - empty composer visible
   - COS controls/content script visible
   - no bootstrap text present in the composer
7. It remained in that state well beyond ordinary composer/model-picker wait bounds.

For the first stuck replacement, I verified the continuation was still pre-Send and explicitly exercised the safe `destinationLost` path. The bridge immediately released that lease and opened a fresh replacement, proving the #141 bridge path no longer waits for the old 15-minute lease timeout.

The second fresh replacement reproduced the same condition:

```text
state = claimed
destinationSend = not-attempted
composer = empty
```

No `destinationAttempt`, `destinationDispatch`, or bootstrap send occurred.

I then reloaded only that exact replacement tab. Because the durable destination state was still `not-attempted`, the new document safely took over the same resume command. Within seconds:

```text
state = committed
destinationSend = sent
command queue = empty
```

The session moved to the new conversation normally and its current-chat context dropped from roughly 600k local estimated tokens to ~9.6k.

[...]

PR #224, opened 2026-09-14
fix: harden Compact & Resume handoff races

K4viar:
## What changed

Compact & Resume has two independent race windows that can strand the continuation during the A → B handoff:

- Project entry can observe the native Project link before ChatGPT's SPA is ready to consume the click. `enterProject()` currently spends one click and gives hydration only 12 seconds, so a swallowed click ends as `ChatGPT could not open the source Project through its native link; nothing was sent`. Give source hydration up to 45 seconds and allow at most 3 native-link clicks, revalidating the exact source chat, project link, composer readiness, attachments, user interruption and command ownership before every retry. All retries still share one bounded 12-second post-click navigation window.
- The resume gate remains valid for 60 seconds, but the recorder waits only 5 seconds before it may create a local session for an unknown replacement chat. Align the recorder wait with `RESUME_CLAIM_WINDOW_MS`, so a legitimately pending A → B commit cannot lose the destination to a shadow session and later fail with `the replacement chat already belongs to another local session`.

Addresses #212 and #218. #167 appears to be a separate redeem/service-worker boundary and is intentionally not claimed as fixed here.

## Validation

- [x] Existing Project-entry timing expectation updated for the 45-second hydration budget; no new regression test cases added.
- [x] `npm run verify` passes. On this pt-BR Windows environment it finishes with 4 failures in `test/renderer-usage.test.ts` because the UI correctly formats USD decimals with commas while those assertions require dot-decimal strings. Running that file unchanged on `origin/main` reproduces the same 4 failures. The full run otherwise reports 4,336 passing tests.
- [ ] Packaging/runtime smoke was run when the change can differ after bundling.

Focused validation passed:
- `npm run typecheck`
- `npx vitest run test/chatgpt-dom-input.test.ts test/bridge.test.ts test/continuation.test.ts` — 532 tests passed


totec448-spec:
The recorder admission fix is already incorporated: a replacement waits within the existing continuation claim window, with delayed-commit, cancellation and expiry regressions. The repeated native Project clicks and expanded navigation wait were intentionally excluded. #212 remains the tracker for the unresolved Project hydration case. Your incorporated contribution is retained in CONTRIBUTORS.md.

Reviewed head: `34c74ffd4a2a`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #229, opened 2026-09-14
Retire app-owned worker tabs after durable stop

### LINKED ISSUE #226: Agents: close app-owned worker tabs immediately after durable stop
## Problem

When a worker finishes its current work and the broker has durably accepted that stop, the browser tab that Chat On Steroids opened for that worker can remain open even though the worker is no longer executing.

That leaves stale worker tabs consuming browser/UI resources and makes long multi-agent runs harder to reason about. The durable worker identity/history should survive, but the browser tab is ephemeral runtime state and should not outlive a completed stop without a reason.

Upstream already has safe tab-close logic for narrow temporary helper tabs, and worker lifecycle already distinguishes `sleeping`, `finished`, `failed`, `waking`, `active`, `detached`, etc. The missing piece is a worker-specific close path that acts only after the authoritative broker stop is accepted.

## Proposed behavior

After a worker stop/finish transaction crosses its existing durable acceptance boundary:

1. determine whether the stopped worker still has an app-owned browser tab bound to its exact conversation/document;
2. if the stop leaves the worker `sleeping`, close that app-owned tab immediately after durable acceptance; the worker conversation/history remains reusable and a later wake reopens the same conversation;
3. if the worker is terminal (`finished`/`failed`), close its app-owned tab after durable acceptance as well, while retaining the broker/session history and retired-worker identity fence;
4. do not close user-owned tabs or a tab whose ownership/document identity is ambiguous;
5. do not let tab closure itself change worker lifecycle state or become evidence that the server-side turn ended.

The close should be a post-commit browser side effect, not part of the authority decision that makes the worker sleeping/terminal.

## Ownership / safety constraints

- Exact conversation + app-owned tab/document proof is required; local worker id alone is insufficient.
- Never close a user-created tab simply because it displays the same conversation.
- Never close while the worker is `active`, `detached`, `invited`, or `waking`.
- A stale close command from an older run/incarnation/document must fail closed.
- The close side effect must happen only after the broker's stop/finish state is durably accepted; if persistence fails/rolls back, the tab stays open.
- Closing the tab must not delete conversation lineage, session history, inbox/queue, context metadata, project binding, or the retired-worker fence.
- A sleeping worker that is messaged later must reopen the same conversation through the existing wake path rather than creating a different worker identity.
- Do not infer completion from browser absence; completion authority remains the broker/turn evidence.

## Suggested regression coverage

- durable worker finish -> exact app-owned worker tab closes once
- durable sleeping stop -> exact app-owned tab closes, worker remains `sleeping` + `revivable`
- persistence failure/rollback -> no close request
- active/waking/detached/invited worker -> no close
- user-owned duplicate tab for same conversation -> only app-owned tab may close
- stale run/incarnation/document close request -> ignored
- already-closed/missing tab -> idempotent success/no harmful retry loop
- later wake of the same sleeping worker preserves worker id/conversation/history and reopens that conversation
- terminal worker tab closure does not remove retired-worker conversation fencing

## Non-goals

- deleting worker conversations/history
- treating tab closure as proof a turn ended
- closing arbitrary ChatGPT tabs
- changing worker stop/finish semantics
- auto-clearing terminal worker rows

A working fork already explored immediate worker-tab retirement. If this direction fits upstream, I can adapt only the smallest post-commit tab-close path and its regression tests to current `main`.

Inmerson:
Fixes #226

## Root cause

The broker already knows exactly when a worker has stopped, and the extension already has the safe physical-close path: exact app-owned document ownership, `clf-tab-close-check`, fresh navigation/document rechecks, pinned/draft guards, and fail-closed behavior.

The gap is the policy handoff between them. `/status` immediately includes published stopped workers in `managedConversations`, but `closableConversations` only reaches those chats later through the generic idle clocks. A worker that has already crossed its durable stop boundary can therefore leave its app-owned background tab around for minutes even though no worker work remains in that tab.

## Change

- Reuse `closableWorkerConversations(0)` as the authoritative list of already-published stopped worker conversations and expose those open, unprotected conversations immediately through `closableConversations`.
- After a browser-observed worker final crosses the existing critical persistence barrier and commits, wake the existing browser maintenance channel so it reads the new policy immediately.
- Do the same after a newly accepted explicit `agents action=finish` commit; idempotent repeated finishes do not publish another wake.

The wake is transport-only. It does not close a tab itself and grants no lifecycle authority. `/status` re-reads current broker state, and the extension still requires exact document ownership and its existing close proof before `chrome.tabs.remove`.

## Safety / invariants

- No close authority exists before `staged.commit()`; persistence failure/rollback leaves the worker active and the tab non-closable.
- No new worker state, timer, ownership registry, polling loop, or recovery authority is introduced.
- Sleeping workers remain `sleeping` + `revivable`; conversation identity/history is retained for the existing wake path.
- Terminal workers retain their broker/session history and existing retired-worker fencing.
- User-owned, stale, ambiguous, pinned, draft-bearing, or navigation-changed documents still fail closed in the existing extension path.
- The extension's existing selected-page reading protection is unchanged.
- Tab absence remains non-authoritative; completion still comes from the broker/turn evidence.

## TDD / validation

On unmodified `f51acbccdd734f524799ea92bb747be765fba1e4`, the focused regression reached the intended boundary: the committed sleeping and terminal cases both failed because `closableConversations` remained empty, while the rollback case already passed.

With this patch:

- `test/worker-tab-retirement.test.ts`: 3/3 passed
- `test/extension.test.ts`: 155/155 passed
- `test/agents.test.ts`: 153/153 passed
- `npm run typecheck`: passed

The branch is one noreply-authored commit directly on current upstream `main`. Full repository verification is left to the upstream PR CI.

## Scope

Three files only: the bridge tab policy / browser-final post-commit wake, the explicit MCP finish post-commit wake, and one focused regression test. No extension close logic or worker stop semantics are rewritten.


totec448-spec:
Closing the immediate-close policy change. Sleeping workers are reusable and the current browser policy intentionally retains a short warm-page interval before idle cleanup. Treating every sleeping worker as immediately closable bypasses that ownership/grace policy. This is a policy decision, not a claim that the proposed branch was merged.

Reviewed head: `07b537dcdec1`. Compared with published main `c5ab88714d3b` (2.1.14).

Issue #42, opened 2026-08-31
Add Chromium CDP detached instance browser extension and headless control or Codex in-app browser extension


winnal:
### Problem or workflow

Right now, it uses your own personal browser instance and the extension to read the chat tabs, to do various operations, sometimes it will bug out if you switch to different tab, or interrupt your control when switching/opening tabs, etc.

### Proposed change

Add a chromium separate browser instanced extension and headless control using CDP, so it can operate in a separate browser instance. It can bind a profile for persistent login. Similar to this other plugin: https://github.com/pauljunsukhan/codex-chatgpt-pro-plugin (it's a good reference for other features too)

### Alternatives considered

Add support for the Codex in-app browser use instead, enabling the extension on there so it runs its own browser instance/profile.

winnal:
After some additional digging, it's not truly headless. We tried using headless and it doesn't work. But it works fine if you can just run it in the background with a detached chrome profile instance and extension for CoS.

winnal:
> After some additional digging, it's not truly headless. We tried using headless and it doesn't work. But it works fine if you can just run it in the background with a detached chrome profile instance and extension for CoS.

This is wrong. It's just extension-free CDP-native CoS operation in a detached managed chrome profile. Basically you can have the personal browser with extension, while the actual prime agent would be able to operate completely detached in its own managed browser instance that is run completely through CDP instead. Will provide more info/create PR when final implementation lands.

totec448-spec:
Thanks for the idea. A separate CDP or embedded-browser path would add another browser lifecycle and login flow for us to maintain. We are keeping one browser transport for now, so I am closing this architecture proposal as not planned. Specific cases where CoS steals focus or acts on the wrong tab are still worth reporting separately so we can fix the actual behavior.

PR #237, opened 2026-09-14
Embed ChatGPT in app-owned Chromium while preserving the existing bridge

### LINKED ISSUE #236: Proposal: host the existing companion bridge in app-owned Electron Chromium
## Problem

Chat On Steroids currently depends on a user-installed Chrome/Edge/Brave process/profile for ChatGPT orchestration. Even when app-created chats are intended to stay in the background, their lifetime and placement still belong to an external browser process, and cold-start/recovery can cross an OS browser/profile boundary.

This is a follow-up to #42, but with a different architecture. #42 proposed a separate CDP/managed-browser transport; this proposal deliberately **does not add a second browser transport**.

## Proposed architecture

Keep the existing companion-extension/bridge protocol and host its ChatGPT documents inside Electron's own Chromium:

- Use one persistent Electron `Session` for ChatGPT login/cookies/storage.
- Load the same bundled MV3 companion extension into that session.
- Represent every logical ChatGPT tab as a real `WebContentsView`.
- Expose a small `chrome.tabs`-shaped compatibility host (`query/get/create/update/remove/reload` plus lifecycle events) to the existing extension background worker.
- Preserve the existing request-id, `sender.tab.id`, `sender.documentId`, navigation-epoch and recovery semantics. No active/recent/visible-tab fallback is introduced.
- Keep background tabs alive with real non-zero layout geometry, mounted but parked offscreen; dock visibility is independent from browser lifetime.
- Keep `window.open`/OAuth child documents in the same persistent Electron session.
- Prewarm one hidden ChatGPT document at app startup so the companion can pair before the user opens the dock.
- Remove the external Chrome/Edge/Brave selector and OS browser-launch path for app-controlled ChatGPT work.

The visible UI is only a resizable dock with pill-style logical tabs. Hiding it must never destroy, suspend or detach the live ChatGPT documents.

## Compatibility intent

The goal is to change the browser host while preserving the existing orchestration contract. Workers, Goal/Loop, plugin refresh, recovery, model selection, Compact & Resume and session attribution should continue to use the same bridge semantics they use today.

A working implementation and deterministic regression coverage are prepared on a fork; the PR will document the host/lifecycle details and validation matrix.

Haz4rdovisk:
## What changed

Closes #236.

This replaces the OS-browser launch/placement path used for ChatGPT orchestration with an **app-owned Chromium host built on Electron `WebContentsView`**, while deliberately keeping the existing companion extension, bridge, attribution model, recovery rules, session semantics and agent orchestration.

image

image

This is also a follow-up to #42. The important architectural difference is that this PR does **not** add a CDP transport or a second browser protocol. The MV3 companion remains the browser-side authority. The change is underneath it: Electron now provides the logical tabs/windows that the existing extension expects instead of delegating those tabs to an installed Chrome/Edge/Brave process.

### Root problem

The previous design had two concerns coupled together:

1. the CoS extension/bridge protocol that owns exact ChatGPT document identity, delivery, recovery and orchestration; and
2. the external browser process/window that happened to host those documents.

That meant app-created ChatGPT work still crossed an OS-browser/profile boundary. Cold start could require launching or locating a user-installed browser, background workers belonged to an external browser window, and app-controlled tabs could still interact with focus/window placement outside the CoS shell.

The smallest durable boundary I found was to keep (1) and replace only (2).

## Architecture after this change

```text
Chat On Steroids BrowserWindow
├─ app renderer/sidebar/composer
└─ app-owned Chromium session: persist:cos-browser
   ├─ WebContentsView: prime/new chat
   ├─ WebContentsView: worker/recovery chat
   ├─ WebContentsView: Compact & Resume successor
   └─ ...

Bundled MV3 companion extension
        │
        ├─ existing bridge/status/input/recovery protocol
        └─ small /browser-host compatibility surface
             query / get / create / update / remove / reload / events
```

There is still one browser-side orchestration implementation: the existing extension. `WebContentsView` is the host for its documents, not a competing controller.

### 1. Persistent internal Chromium session

`src/main/internal-browser.ts` owns a persistent Electron partition (`persist:cos-browser`). The same session is used by every logical ChatGPT tab, so login state, cookies, local storage and the bundled companion extension are shared across prime chats, workers, recovery tabs and resumed chats.

The bundled MV3 extension is loaded into that session before normal browser work begins. The internal session removes Electron's product token from the user agent but otherwise leaves Chromium/provider behavior intact.

The external `ChatGPT browser` preference (`chrome` / `edge` / `brave`) is removed because app-controlled ChatGPT work no longer selects an OS browser.

### 2. One real `WebContentsView` per logical ChatGPT tab

Each logical tab is a separate `WebContentsView`; this is not a DOM iframe and workers do not share one document.

The host keeps Chrome-like metadata for each view (`id`, active state, pending URL, title, status, last access, opener, pinned/user-held semantics). `webContents.id` is the logical tab id exposed to the companion, and the host presents one logical window id because the app owns one Chromium tab family.

The compatibility host implements the subset the extension needs:

- `tabs.query`
- `tabs.get`
- `tabs.create`
- `tabs.update`
- `tabs.remove`
- `tabs.reload`
- a bounded lifecycle-event journal with host generation/cursor semantics

That surface is exposed through the existing loopback bridge as `/browser-host`. URLs accepted for tab creation/navigation are restricted to ChatGPT origins.

The extension's higher-level orchestration logic still reasons in terms of tabs/documents. It does not need Electron-specific branches spread throughout worker/recovery/input code.

### 3. Hidden does not mean detached

This was the most important lifecycle detail found during manual testing.

An early implementation detached hidden `WebContentsView`s from the owner `BrowserWindow`. The page stayed alive enough for basic navigation, but ChatGPT's model picker relies on real layout geometry (`getClientRects()`). A hidden worker/input could therefore be claimed by the extension and then stall during browser preparation until timeout.

The final behavior is:

- once the owner `BrowserWindow` exists, **every live `WebContentsView` remains mounted**;
- the selected tab receives the real dock bounds only while the dock is open;
- every background tab, and the selected tab while the dock is hidden, stays `setVisible(true)` but is parked fully offscreen at negative `x`;
- parked views keep a real viewport of at least `420 x 720`, or the current dock size when larger;
- `backgroundThrottling` is disabled;
- hiding the dock only parks views; it does **not** call `removeChildView`;
- views detach only for actual tab retirement, owner-window replacement/teardown, or browser shutdown.

This makes a hidden CoS tab behave much closer to a real background Chrome tab: it is not user-visible, but layout-dependent provider UI continues to exist and automation can interact with it normally.

### 4. Dock and tab UI are presentation only

The renderer adds a resizable ChatGPT dock with a thin pill tab strip.

- The dock can be shown/hidden without changing browser lifetime.
- Pills switch the active `WebContentsView` instantly.
- Closing a pill retires only that logical tab.
- Background worker/recovery tabs remain alive while another tab is displayed.
- The preferred dock width is local renderer state.
- Explicit user opens can request the dock; background-created work does not.

The renderer never embeds the remote page itself. It only reports the slot bounds over IPC; the main process positions the native `WebContentsView` there.

### 5. Startup/prewarm no longer launches an OS browser

The bridge and internal Chromium are initialized before the desktop window is considered ready. Startup loads one real hidden `https://chatgpt.com/` document so the bundled companion can pair/wake without requiring the user to click the dock first.

`browser-startup.ts` now has one cold-start owner for this internal host. If the companion wake channel already exists, normal extension election continues to own placement. If it does not, the internal host opens the requested marked document; there is no process probe or `shell.openExternal` browser launch fallback for ChatGPT orchestration.

A development-only `COS_DEV_USER_DATA`/`COS_DEV_EXTENSION_DIR` seam lets an unpackaged dev instance use a disposable Electron profile beside an installed CoS instance. Packaged builds ignore that escape hatch.

### 6. `window.open` / OAuth keep browser semantics

Remote `window.open` is handled through Electron's `createWindow` callback by creating another hosted `WebContentsView` in the same persistent session. That preserves opener/session behavior needed by provider flows without creating a separate native browser window.

Non-web `mailto:`/`tel:` targets can still go through the OS; app-controlled ChatGPT navigation remains inside the internal host.

### 7. Extension adaptation preserves identity instead of replacing it

The extension now routes the tab operations it used to perform through `chrome.tabs` to the `browserTabs` compatibility object backed by `/browser-host`.

The important part is what was **not** relaxed:

- request-id correlation is unchanged;
- `sender.tab.id` remains exact;
- `sender.documentId` remains exact;
- navigation epoch/document lifecycle checks remain exact;
- there is no active/recent/visible-tab fallback;
- unattributed calls still fail closed;
- workers still have separate conversations/documents;
- recovery still depends on exact browser/page evidence.

The bridge protocol is bumped from 13 to 14 because the app and companion now need the browser-host request/response shape together.

### 8. Chrome lifecycle semantics are emulated deliberately

Electron's navigation events are not a 1:1 match for `chrome.tabs.onUpdated`.

During testing, treating every `did-start-navigation` as `status: loading` caused redirects/restarts within one real load cycle to look like multiple document boundaries. A replacement content-script document could register and then be terminalized by a later navigation event from the same spinner cycle, producing intermittent `WORKER_IDENTITY_LOST`.

The host now maps lifecycle as follows:

- `did-start-loading` => the single `status: loading` document boundary;
- `did-start-navigation` => URL bookkeeping only (main-frame, non-in-place navigation);
- `did-stop-loading` => `status: complete`;
- same-document navigation updates the URL without fabricating a loading boundary.

The host also journals tab update/remove events so an MV3 service worker restart can resume from a cursor/generation instead of losing lifecycle changes.

### 9. User-held/pinned semantics protect visible chats

The existing extension already treats Chrome-pinned tabs as user-owned and avoids recycling them. The internal host mirrors that semantic: a tab the user explicitly opens/views can be exposed as logically pinned/user-held even though Electron has no native tab strip.

That prevents opening another recorded chat from letting maintenance recycle the chat the user just left.

### 10. Compact & Resume recovery is tied to exact command identity

The first integration pass removed an older recorder-local five-second cutoff and made unknown replacement observations wait on the continuation's existing bounded gate. Post-integration recovery testing then exposed a second, longer race in real history.

The recorded sequence was:

- CoS opened the replacement at `00:53:16.719`;
- ChatGPT materialized the distinct replacement conversation at `00:54:17.948` — **61.229 seconds later**;
- the coarse 60-second recorder gate had expired about 1.2 seconds earlier;
- the recorder therefore created a shadow local session for B;
- the later legitimate A -> B continuation commit found its destination already owned and was refused. The UI eventually surfaced the misleading `Failed — the replacement chat is not a distinct conversation` even though A and B were different ChatGPT conversation ids.

Simply increasing the global timeout would only move that race and would make unrelated New Chat recording wait longer. The final path instead uses the exact one-shot authority the resume already has:

- only after the destination send is durably armed does `content.js` retain the exact random `resumeCommandId` redeemed by that document;
- before request-id ownership is sent, the page also proves that the current provider transcript contains the matching `[[CLF-RESUME:<token>]]` marker;
- the current MV3 document forwards that command id with the existing `/correlations` request;
- the bridge accepts it only if the command still exists as a claimed resume, still owns that continuation via `continuationClaimedBy`, and the destination send is already `dispatched-unresolved` or `sent`;
- the A -> B session commit happens **before** request attribution is allowed to initialize B;
- retryable/refused commit states return the existing retryable 503 path instead of creating a second session;
- the normal command ACK / stable marked-message paths remain the terminal authority and can still win the same transaction first.

There is still no active/recent/visible-tab fallback. A normal unrelated chat carries no resume command provenance, so once the bounded compatibility gate expires it records immediately rather than inheriting the continuation's longer lifetime.

This keeps the original session ownership checks intact while eliminating both the old 5-second shadow-session race and the observed >60-second race.

### 11. Legacy external-browser placement code is removed

The old browser selector/launcher and background-window layout implementation are removed together with tests that only described that architecture. Their coverage is replaced by internal-browser host tests plus existing bridge/extension/input/recovery tests adapted to the new host contract.

### 12. Refreshed onto upstream 2.1.13 without restoring removed behavior

The PR branch was rebuilt as **one feature commit directly on upstream `2f9be7e` (2.1.13)** rather than merging the old 2.1.11 implementation forward. That made the conflict resolution reviewable file-by-file and keeps later upstream work in its native form.

The refresh preserves the current 2.1.13 Skills/text-skill setup, release/setup reliability, current model-picker behavior, input/recovery fixes and session semantics. Internal Chromium adaptations were applied around those owners instead of reviving stale 2.1.11 browser/session logic. The companion/app protocol remains 14 and all browser-host changes continue to fail closed on exact document/request identity.

## Behavior intentionally preserved

This PR changes where ChatGPT documents live, not how CoS decides who owns work.

The following flows keep their existing semantic owners and were exercised against the new host:

- normal/new-chat input delivery;
- model discovery/model picker and reasoning selection;
- durable session recording;
- exact request attribution;
- worker spawn, wake, reuse, finish and sleeping history;
- browser wake/recovery;
- Goal automation;
- Loop/session-finish machinery;
- plugin refresh;
- project successor/workspace flows;
- native attachments/input replay;
- Compact & Resume and handoff history.

No CDP controller, browser automation fallback, iframe multiplexing or shared worker chat was added.

## User-visible behavior

Before:

- CoS chose Chrome/Edge/Brave for app-originated ChatGPT launches.
- Background orchestration lived in an external browser process/window/profile.

After:

- CoS owns a private persistent Chromium session.
- The user can sign in to ChatGPT in the integrated dock.
- App-controlled ChatGPT tabs start with CoS and can remain fully alive while the dock is hidden.
- Multiple chats/workers appear as pill tabs when the dock is opened.
- Closing/switching a pill only affects that internal tab.
- External browser selection is no longer part of CoS ChatGPT orchestration.

### Deterministic automated coverage

The refreshed single-commit branch was validated on upstream `2f9be7e` after the final recovery fix.

Full `npm run verify` result:

- **181 test files passed, 4 skipped**;
- **4457 tests passed, 42 skipped, 0 failed**;
- the separate `test/mcp-shutdown.test.ts` verify step: **2/2 passed**;
- public-history privacy check: passed;
- third-party notices/native-source checks: passed;
- TypeScript typecheck: passed;
- Electron load check: passed.

The final focused browser/recovery gate also passed:

- `test/content-script.test.ts`
- `test/extension.test.ts`
- `test/bridge.test.ts`
- `test/continuation.test.ts`
- **1223/1223 tests passed**.

`npm run build` passed and `git diff --check` passed.

New/updated deterministic coverage includes:

- hidden owner attach => real offscreen bounds;
- background tab creation while the dock is hidden;
- show => selected exact dock bounds / background remains parked;
- hide => selected becomes parked and no `removeChildView` occurs;
- retirement/owner replacement/shutdown detach semantics;
- one Electron load cycle => one Chrome-like loading boundary even across multiple navigation events;
- URL-only same-cycle navigation does not terminalize the replacement extension document;
- delayed Compact & Resume replacement cannot create a shadow session inside the ordinary gate;
- a replacement appearing **after the old 60-second gate** can still move the exact original session when its current document supplies the exact redeemed resume command and matching continuation marker;
- an unrelated New Chat without that provenance still receives its own session immediately rather than waiting on the continuation lifetime;
- an early resumed request-id can be correlated before the normal command ACK without relaxing exact ownership.

### Manual runtime / E2E validation

The implementation was exercised in a real Electron dev instance with the bundled extension and real ChatGPT documents, not only mocks.

#### Hidden input delivery

The original detached-view failure was reproduced first: the extension could claim an input, then browser preparation timed out while the dock was hidden. With views mounted and parked offscreen, a clean hidden attempt was claimed in ~129 ms, acknowledged in ~1.1 s and durably marked sent without opening the dock.

#### Agents

Two existing workers were reused with the dock hidden (no fresh worker substitution). Both background chats reopened in separate hosted views, returned distinct acknowledgements, called worker finish and returned to sleeping state. No `WORKER_IDENTITY_LOST` occurred. Reopening the dock showed the worker tabs still alive with their conversation content intact.

#### Tab switching/closing

Prime/worker/test tabs were switched through the pill strip. Background views remained alive. Closing one disposable pill retired only that view and immediately exposed another still-live tab; unrelated tabs/history were preserved.

#### Compact & Resume

A disposable conversation was compacted into a replacement conversation through the internal browser host. The handoff reached the configured 50,000-character bound, was stored durably, the replacement tab was created, and the **same local session** was rebound from chat B to chat C. The continuation committed without creating a shadow local session, and the first request from C was attributed to C exactly.

#### Goal with dock hidden

A disposable Goal session produced an initial response, then the dock was hidden. Goal generated and sent the continuation through the parked view, received the requested completion, reevaluated the chat and logged `goal is met ...; nothing was sent`. Reopening the tab showed `Goal reached` with the same history.

#### Loop caveat

Basic Loop input/tool activity was exercised with the dock hidden. A full manual `Only finish` continuation was not possible in this environment because `session_finish` is an Astra-only tool and is not callable from an ordinary chat/composer. The Loop/session-finish path remains covered by deterministic tests, including `session-finish.test.ts` 28/28.

#### Restart/persistence

The dev Electron process was restarted multiple times during the original internal-browser validation, including prewarm/session recovery.

After the 2.1.13 refresh and the exact resume-correlation fix, recovery was repeated against an isolated copy of the real local catalog containing **85 recorded sessions**. Three older sessions were switched repeatedly A -> B -> C -> A, then the dev app was closed through its normal window shutdown, reopened on the same cloned state, and the same sessions were selected again. Titles and historical transcript content returned correctly each time; the catalog remained **85 sessions** before and after restart, and no shadow/phantom session was created by the navigation/reopen exercise.

That final 85-session pass intentionally used an isolated local profile and validates local session/catalog persistence and renderer restore. It is not claimed as a fresh authenticated ChatGPT Compact & Resume E2E run; the exact >60-second provider race itself is covered deterministically above from the real recorded timing.

## Compatibility / migration notes

- The refreshed PR is exactly one feature commit on upstream `2f9be7e` / app 2.1.13.

- Existing `ui.chatBrowser` configuration is no longer used; the UI selector and schema field are removed.
- Integrated Chromium uses its own persistent profile, so users may need to sign in there even if already signed in to an external browser.
- Companion/app protocol moves to version 14; a matching bundled companion is expected.
- The app still restricts extension-created/navigated host tabs to ChatGPT origins.
- The change uses Electron `WebContentsView` from the Electron version already pinned by the project; no new runtime dependency is added.

## Validation checklist

- [x] Added or updated deterministic regression tests where behavior changed.
- [x] `npm run verify` passes: 181 files passed / 4 skipped, 4457 tests passed / 42 skipped / 0 failed, plus `mcp-shutdown` 2/2 and all non-test verify steps.
- [x] Typecheck, Electron load, privacy/notices/native-source verification and `git diff --check` pass.
- [x] Production `npm run build` passes.
- [ ] Packaged-binary smoke was not rerun for this refresh; the browser/session behavior was exercised in Electron dev runtime and the PR adds no new runtime dependency.

## Review pointers

For reviewers who want the shortest path through the implementation:

1. `src/main/internal-browser.ts` — ownership/lifetime/layout of hosted tabs and the Chrome-like host surface.
2. `extension/background.js` — `browserTabs` compatibility adapter and event synchronization; higher-level attribution/recovery remains in place.
3. `src/main/index.ts` + `src/main/browser-startup.ts` — startup/prewarm and removal of OS-browser launching.
4. `src/renderer/internal-browser.ts` — dock/pill presentation and bounds IPC.
5. `src/main/session/resume-gate.ts` + recorder call site — event-driven replacement-chat gate.
6. `test/internal-browser.test.ts`, `test/extension.test.ts`, `test/bridge.test.ts` — the lifecycle/identity/race regressions.


Haz4rdovisk:
[internal-chromium-architecture.md](https://github.com/user-attachments/files/32211974/internal-chromium-architecture.md)

PR #243, opened 2026-09-15
feat(i18n): add Traditional Chinese locale

aliceric27:
## What changed

Added Traditional Chinese (Taiwan) (`zh-TW`) as a supported UI language alongside English and Simplified Chinese.

The existing localization layer only supported `en` and `zh-CN`, so Traditional Chinese could not be selected, persisted, or rendered independently. The change extends the existing language/catalog mechanism rather than introducing a separate localization path.

This PR:

* Adds a complete `zh-TW` catalog matching the existing `zh-CN` source keys and interpolation placeholders.
* Adds `繁體中文` to both the initial Setup language selector and Settings → Appearance → Language.
* Persists and restores `zh-TW` through the existing `cos.ui.language` preference.
* Falls back to English for unsupported saved locale values.
* Uses Traditional Chinese font fallbacks and allows the language selector to wrap at narrow widths.
* Updates renderer tests and Chromium layout verifiers to cover `zh-TW`, including language switching, persisted preferences, mounted labels, permission guidance, Skills, Timeline, setup/profile UI, dropdown layout, and label repaint behavior.
* Keeps user-authored/provider content unchanged while translating app-owned UI.
* Updates the renderer/i18n architecture documentation for the new locale.

The Traditional Chinese catalog was also reviewed for Taiwan terminology, replacing Mainland-oriented wording such as `通過` (when meaning “via”), `暫無`, `無響應`, `閾值`, and similar UI terminology with Taiwan-appropriate equivalents.

## Validation

* [x] Added or updated a deterministic regression test where behavior changed.
* [ ] `npm run verify` passes.

  * Not run in this checkout because `node_modules` is not installed and `npm` is not available in the current environment.
* [ ] Packaging/runtime smoke was run when the change can differ after bundling.

  * Not run for the same environment limitation above.
* [x] No unrelated formatting, generated output, local debugging notes, or private data is included.
* [x] Screenshots, logs and examples use placeholders instead of real usernames, paths, chat text, IDs or credentials.
* [x] Security-sensitive details are being handled privately instead of disclosed here.

Additional checks completed:

* `git diff --check`
* JavaScript/CJS syntax checks for the modified Chromium verifier scripts
* TypeScript syntax checks for the modified renderer/test files
* `zh-CN` / `zh-TW` catalog key parity across 1,135 entries
* Interpolation placeholder preservation
* Technical URL preservation
* Unicode/control-character and navigation-symbol integrity checks
* Targeted scan for Mainland Chinese terminology in the `zh-TW` catalog


totec448-spec:
Traditional Chinese, Taiwan terminology and language controls are incorporated in current main and extended for the current interface. Both Traditional Chinese renderer tests passed in this review. Your original contribution remains credited in CONTRIBUTORS.md.

Reviewed head: `7d6b234878a4`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #245, opened 2026-09-15
Enable native macOS fullscreen

igorbelchior86:
## What changed

The main `BrowserWindow` explicitly set `fullscreenable: false` on every platform, which disables the standard macOS green-button fullscreen behavior. Enable `fullscreenable` only on macOS so the window can enter native Space-backed fullscreen, while preserving the existing Windows/Linux behavior.

A regression test locks the platform-specific setting in place.

## Validation

- [x] Added/updated a deterministic regression test for the behavior change.
- [x] `npm run verify` passes: 4,411 main-suite tests plus 2/2 isolated MCP shutdown tests.
- [x] macOS arm64 unpacked package built successfully.
- [x] macOS bundle smoke and packaged-runtime smoke passed.
- [x] Isolated packaged-app runtime check confirmed the main window is fullscreenable, enters native fullscreen, and restores its previous bounds after exit.
- [x] No screenshots/logs containing private user data are included.
- [x] No security-sensitive details are disclosed.


igorbelchior86:
Screenshot 2026-09-15 at 13 40 23

Issue #238, opened 2026-09-14
Requesting model / Reasoning not confirmed


jackbarunz:
### App version

2.1.11

### Operating system and version

Windows 11

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT workspace / custom MCP access

Other or not sure

### What happened

POSSIBLE BUG

<img width="728" height="170" alt="Image" src="https://github.com/user-attachments/assets/f8306128-9f79-4779-ab4c-9c817e11eb92" />
I started a new chat asking to code a landing page (not really important), which lasted around 1 hour total. At some point the chat just couldn't load the models, no matter what I did. I tried refreshing them from the selection bar in the chat, also tried in the settings and even in the chatgpt browser tab connected to the chat. Just wouldn't load them. 

IMPORTANT 1: if I start any new chat, Chat On Steroids works just fine. I think there is simply a problem in the long run. 
IMPORTANT 2: I chose Work instead of Chat just for trying and it worked just fine at the beginning, but like I said, in the long run it stopped loading the models. I don't know if Work selection is what messed it up but I don't think so, since it worked in the beginning.

Hope I found a bug and I'm not simply being clueless

### Anything else

```text

```

totec448-spec:
Yeah that's a bug

jackbarunz:
> Yeah that's a bug

Any known solutions?

jackbarunz:
I've been keeping an eye on updates and the whole repository in general so I'd like to keep helping by saying that even the latest version has problem with model and reasoning. I may have found a pattern which doesn't only concern long chats, but also chats whose browser tab gets closed (For example, on computer shut down). I wonder whether in order for a chat to keep running, its browser tab must stay open

Issue #228, opened 2026-09-14
Plus/Personal: OpenAI Secure MCP Tunnel shows no ChatGPT workspace — what is the supported setup path?


Hoodie2389:
### Context

I'm trying to use Chat On Steroids specifically with a normal **ChatGPT Plus / Personal** account, rather than upgrading to Business just to get the local Core/worker/terminal workflow.

The repo messaging appears to support this use case:

- README: "Your files. Your terminal. Your ChatGPT plan."
- README: "Uses your ChatGPT conversation rather than invoking Codex directly."
- v2.0.7 release notes: "plus = gpt 5.6; pro = astra"

There are also issues where users report **Personal ChatGPT workspace with Core connected** (for example #202), so Personal appears to be a supported configuration in at least some setups.

### What I'm seeing

When I follow the documented **OpenAI Secure MCP Tunnel** setup and go to Platform → Tunnels → Create tunnel:

- **Organizations** shows my `Personal (org-...)` API organization.
- **ChatGPT workspaces** is completely empty.
- Therefore I cannot associate the tunnel with the Personal ChatGPT workspace I actually use.

This is the exact blocker: the API organization exists, but no ChatGPT workspace is selectable in the tunnel form.

### Question

For a **ChatGPT Plus / Personal** user, what is the currently supported way to connect CoS Core?

Is the intended setup:

1. OpenAI Secure MCP Tunnel somehow with the Personal workspace,
2. Cloudflare Quick Tunnel + custom MCP app in the Personal workspace,
3. another setup path,
4. or is full Core currently only possible on Business/Enterprise despite the repo's Plus/Personal positioning?

If Cloudflare Quick Tunnel is the intended Plus/Personal path, it would be very useful to state that explicitly in `docs/setup.md`, because the current first-choice OpenAI tunnel flow leads to a dead end when the ChatGPT workspace dropdown is empty.

### Why this matters

The value proposition of CoS for me is specifically to get a Codex-style local coding/agent workflow out of the existing ~$20 ChatGPT Plus subscription, instead of adding Business seats or moving the whole workload to usage-billed API calls.

I don't want to work around OpenAI plan restrictions or safety controls; I just want to know the officially supported CoS setup for Plus/Personal today.

### Environment

- ChatGPT plan: Plus
- ChatGPT workspace: Personal
- OpenAI API organization: Personal (`org-...`)
- Platform tunnel creation page: Organizations populated, **ChatGPT workspaces empty**

[...]

Issue #187, opened 2026-09-12
Image generation and usage


TheBigBrainChad:
Image generation sometimes fails to recognize an image that is already available in the coding/session environment as a valid reference input.
The failure can occur even when the image is visible, readable, and successfully inspected by the assistant.
The generator may return: there is no usable image target in the conversation.
This suggests the image-viewing/file environment and the image-generation environment do not always share the same attachment/reference state.
A local filesystem path is not enough to guarantee that the image generator can use the image.
There is no clear supported mechanism to explicitly bind a repository/local image file to the generator as its reference image.
Because of that, reference-conditioned generation can fail before generation even starts.
The problem is inconsistent because the image may be accessible to other tools while still being unavailable to image generation.
Image output format is also inconsistent in some generation paths.
Some outputs were returned as opaque WebP files instead of transparent PNGs.
In those cases, the image contained a painted checkerboard background rather than a real alpha channel.
This makes an image appear transparent visually while actually being fully opaque.
The generator therefore does not always preserve requested transparency semantics.
Generated dimensions can also differ from the expected output contract without a reliable way to enforce the exact runtime size.
A separate issue affects saving/downloading generated images.
Sometimes the image is successfully generated but cannot be transferred into the coding workspace.
Artifact download attempts can fail because the generated file URL is rejected by the file-download bridge.
One observed error reported that the generated URL was outside the trusted file host.
This means successful generation does not guarantee that the resulting file can actually be saved locally.
The download failure is intermittent rather than a consistent unsupported-operation error.
Retrying may sometimes work through another path, which makes the behavior difficult to automate reliably.
There is also no stable direct mapping from a generated image result to a guaranteed local artifact path.
As a result, the workflow has two independent failure points: reference-image binding and generated-artifact retrieval.
The first failure prevents generation from using the requested source image.
The second failure can lose access to an image that was already generated successfully.
These issues make image generation unreliable inside coding workflows that depend on existing project images.
The expected behavior should be that any explicitly supplied local/attached image can be deterministically passed to image generation.
The generated result should preserve requested alpha/transparency and requested file format.
The generated artifact should always expose a stable, trusted, directly downloadable/savable file reference.
In short, the core bugs are unreliable reference-image binding, inconsistent transparency/output format, and intermittent failure to download/save generated artifacts.

lavalava45:
I have a concrete reproducible `download_artifact` failure for the saving part of this issue.

Environment:

* Chat On Steroids 2.0.9
[...]

Issue #168, opened 2026-09-11
새 ChatGPT 채팅 전환 중 실제 전송된 desktop input의 ACK가 유실됨


ventianima-lab:
## 증상

Chat On Steroids 2.0.8에서 새 ChatGPT 채팅으로 desktop input을 보낼 때, ChatGPT가 메시지를 실제로 수락했지만 SPA 전환 중 원래 content document가 교체되면 `/input/ack`가 앱에 도착하지 않을 수 있습니다.

앱은 중복 전송을 막기 위해 해당 입력을 재전송하지 않는 점은 올바르지만, 45초 뒤 요청 행을 아래 오류와 함께 `cancelled`로 남깁니다.

`Stopped waiting for delivery confirmation. The message may already have been sent; it will not be resent.`

## 재현 증거

- 앱 로그: input이 browser에 claim됨
- 같은 요청의 `sendAuthorizedAt` 이후 5.4초에 새 대화의 recorder에 정확한 user message가 생성됨
- 그 대화에는 최종 assistant message도 존재함
- durable input 행에는 `conversationId`, `deliveredSessionId`, `messageId`, `deliveredAt`가 모두 없음
- 동일 텍스트를 다시 보내지 않았으며 실제 전송은 1회였습니다

## 기대 동작

새 채팅의 실제 user message가 만들어졌다면, SPA route/document 전환과 경쟁하더라도 exact input owner + conversation + native message receipt를 durable ACK journal에 넘겨야 합니다. ACK가 정말 불명확한 경우에는 지금처럼 재전송하지 않아야 합니다.

## 제안

1. 새 채팅 send receipt를 route/document 전환 전에 durable ACK journal이 인수하거나,
2. 실패 시 앱 recorder의 유일한 exact user row로 동일 입력을 fail-closed 방식으로 복구하십시오.

복구는 ambiguous cancellation, exact authorized-send window, exact authored text, unique recorded row를 모두 요구해야 하며 중복 후보가 있으면 아무 것도 귀속하지 않아야 합니다.

ventianima-lab:
추가 실기기 재현: 새 대화뿐 아니라 이미 알려진 세션에서도 같은 ACK 유실이 발생했습니다.

- control request: `b63c33c3-0a6f-47d0-aa5d-8ce718460354`
- 앱 outbox: browser claimed 후 45초 뒤 동일한 ambiguous cancellation
[...]

PR #246, opened 2026-09-15
Record 47-issue audit and published-code reproductions

### LINKED ISSUE #167: Compact & Resume replacement can stay claimed before destinationAttempt until tab reload

totec448-spec:
The resumed sweep had no canonical `ISSUE_TRIAGE.md`. This adds every one of its 47 open issues with a verdict, implementation boundary, reproduction evidence, disposition and remaining blocker. The sweep remains unfinished: 24 feature requests, two scoped bug reproductions, and 21 unresolved reports. A blocked reproduction is explicitly not counted as a successful negative test.

The evidence directory contains executable, dependency-free characterizations against published commit `2f9be7eb9d6d1f13ccac9019e20965af15d4e4be`:

- #212: source hydration at 11 seconds succeeds, while hydration at 13 seconds fails at the independent 12-second cutoff. Eight timing/guard cases passed.
- #187: all three report-provided regional hosts are rejected before fetch; three accepted-host controls and nine rejection guards passed. This proves the rejection path, not independent native origin for those hosts.

Both reports deliberately say `fixVerified: false`. Source copies and executable/result files were checked against their published Git blob hashes. The fixtures are isolated from the normal product/test build and contain no credentials or live signed file references.

The previous local test reports contain 221 + 2 + 2 actual passes; their 774 skipped/pending cases are not counted as passes. These are retained earlier results, not new validation of today's concurrent edits. The audit also records conflicting current official guidance relevant to #228.

GitHub labels were updated on 23 confirmed feature requests, with documentation labels on two of them; no issues were closed. Core, Desktop and Plugins became unavailable with `tunnel_client_not_seen`, preventing current-worktree integration and signed-in browser checks. The newer dirty Windows worktree is different from this published base and has not been overwritten, reset or claimed as verified. No product code, installation or release change is included.

Keep this PR as a draft until the remaining current-app reproduction and integration work in `ISSUE_TRIAGE.md` is completed. This PR does not resolve #167, #168, #109, #104, #238, #187 or #212.


totec448-spec:
The new published audit accounts for every item in the 65-issue/39-PR inventory and supersedes this older 47-issue draft. The still-unresolved regional artifact-host/transparency case in #187 and Project hydration case in #212 remain explicitly open in that report. Closing the obsolete draft without discarding those findings. Published in #316 (https://github.com/totec448-spec/chat-on-steroids/pull/316), merge commit `962d040b423f`. This is a source integration for the next release; it does not install or publish a new application version.

Reviewed head: `6a9ee5a0fef6`. Compared with published main `c5ab88714d3b` (2.1.14).

Issue #250, opened 2026-09-15
Model discovery expires after browser wake without observation


holzhaker1:
## Context

The 2.1.13 runtime log shows repeated `model discovery requested` and `model discovery browser wake completed` events, but no `model discovery observed` event before the 120-second request expiry. The same run also reports the tunnel-client control-plane messages `dispatcher received unsupported channel` and `unsupported channel "harpoon"`.

## Expected behavior

- A browser wake should either complete the model-catalog observation or produce a bounded, actionable failure that can be retried; it should not leave the app waiting until an unexplained expiry when the browser handoff has already started.
- Known benign Harpoon control-channel incompatibility should not be presented as an app/tunnel warning, while other unsupported-channel and tunnel failures must remain visible.

## Acceptance criteria

- Add regression coverage for the model-catalog handoff at the extension/background-to-content boundary, including the worker-lifecycle or ownership path that can lose the observation.
- Preserve exact nonce/tab/document checks and do not publish a partial model catalog.
- Add narrow regression coverage for the two benign Harpoon messages and retain diagnostics for unrelated tunnel errors.
- Verify the normal model refresh flow on a real signed-in ChatGPT page and keep the local app usable.

totec448-spec:
The browser-wake discovery custody problem and scoped Harpoon diagnostics were adapted from #251 into current main. Discovery now retains its exact tab/document/navigation owner across the relevant lifecycle. Closing this specific implemented correction; separate current model-discovery reports remain open.

Checked against published main `c5ab88714d3b` (2.1.14).

PR #251, opened 2026-09-15
Fix model discovery expiry after browser wake

holzhaker1:
## Summary\n\n- persist model-discovery tab custody across MV3 service-worker restarts\n- require the exact tab URL, document, and navigation epoch for observations\n- keep startup discovery passive and distinguish wake dispatch from observation completion\n- suppress only the known Harpoon unsupported-channel diagnostics\n\n## Verification\n\n- focused model-discovery, window-lifecycle, tunnel, and chat-model tests pass\n- typecheck, build, privacy, and notices checks pass\n- full suite: 4,488 passed; remaining failures are unrelated Windows/locale environment issues\n\nCloses #250


totec448-spec:
The model-discovery tab/document custody, startup handling and Harpoon diagnostics were adapted into current main. The Harpoon filter now requires exact channel evidence rather than searching arbitrary serialized event data. This closes the incorporated patch, not every separately reported model-discovery problem. Attribution is retained in CONTRIBUTORS.md.

Reviewed head: `762976acd6b0`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #254, opened 2026-09-15
feat: add composer slash mode panels

igorbelchior86:
## What changed

Describe the root cause and the smallest behavior change that fixes it.

## Validation

- [x] `npm run verify` passes.
Screenshot 2026-09-15 at 18 59 32


Haz4rdovisk:
Does it support user imported skills?

igorbelchior86:
> Does it support user imported skills?

Not yet, only built-in commands for the moment.

totec448-spec:
Closing this composer/mode replacement. It can replace an independently authored composer draft and use the current attachment state for a different submitted draft; mode recovery also depends on a capped outbox history. Current Skills/slash invocation is already available without introducing these draft-ownership problems.

Reviewed head: `9f40d72f007d`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #263, opened 2026-09-16
fix: reconcile agent plan lifecycle

igorbelchior86:
## What changed

Agent Plan state could remain visibly stale after the actual turn had finished because `update_plan` was durable session state with no authoritative turn lifecycle. A later reconciliation could also loop when an accepted `update_plan` created a fresh revision before the browser had supplied `activeTurnId`.

This change gives the stored Agent Plan an explicit lifecycle:

- unfinished plans pause on a proven completed turn instead of remaining live forever;
- normal later turns do not silently resurrect paused progress;
- a later `update_plan` explicitly reconciles/reactivates the plan;
- paused-plan context is supplied to follow-up work and MCP results without altering visible user text;
- identical `update_plan` payloads are idempotent and do not create another revision;
- one accepted reconciliation consumes the reminder even when turn identity arrives late;
- `session_finish` holds while unfinished Agent Plan steps still need reconciliation;
- legacy plan files remain readable and derive lifecycle state from recorded turn boundaries.

## Validation

- [x] Added deterministic regressions for stale-plan retirement, multi-turn reconciliation, identical update idempotency, and the late-`activeTurnId` reconciliation loop.
- [x] `npm run verify` passes on top of current `origin/main`.
- [x] 4,542 tests passed (4,540 main-suite + 2 MCP shutdown), with 129 skipped and 0 failures.
- [x] Typecheck, privacy checks, third-party notices, and native-source verification pass.
- [x] No unrelated generated output, local debugging notes, credentials, or private data are included.


totec448-spec:
Closing this plan-lifecycle rewrite. A completion from another turn can pause a newer plan without matching activatedByTurnId, and unfinished presentation steps can hold finish independently of actual execution. Plans remain a projection of work rather than a second execution/finish authority.

Reviewed head: `80ef2e31e3d8`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #267, opened 2026-09-16
fix: recover chats whose tab Chrome discarded or froze

AndersonBY:
## Root cause

Chrome can shelve an app-used background tab two ways: Memory Saver **discards** the document and Energy Saver **freezes** its timers. A tab in either state keeps its URL and answers every `chrome.tabs.query`, so none of the app's existing recovery paths ever fired for it:

- `/closed` never arrives (the tab was never closed), and the shell counts as "open" in the extension's `openConversations` report, so the missing-tab path was bypassed;
- the silence sweep is gated by `tabRecoveryWanted()` and never runs for ordinary chats at all;
- a deferred worker revival pinged the dead page and read the failure as "the exact tab is still there", so no replacement was allowed — the revival then sat until its 90-second deadline failed it.

Meanwhile the page is gone or suspended: recording stops, and queued input / Goal continuations / wakes cannot land in it.

`autoDiscardable` protects only against discarding; no extension API exempts a tab from freezing, so the answer is detection and repair rather than prevention.

## Change

**Extension (`background.js`)**
- The maintenance pass reports discarded/frozen shells separately as `stalledConversations` in the `/status` POST (the tab query already returns those flags; `frozen` exists on Chrome 132+ and degrades to absent before it). `openConversations` keeps its exact old meaning.
- `recoverDeferredRevivals` reloads a discarded exact tab and lets the reloaded document's registration re-enter the flow, instead of parking the revival on a dead page until its deadline.
- Tabs created for an input are marked `autoDiscardable: false` at creation, and the `cos-input` marker now counts as an in-flight opening in the protection pass — closing the window before the conversation binds and the app's policy set takes over.

**Bridge (`bridge.ts`)**
- `/status` validates the optional `stalledConversations` list (subset of the reported open conversations) and runs each entry through a new `queueStalledTabRecovery`: deliberately the missing-tab decision minus the close side effects — the shell stays open, no run ends, and the service worker reloads the exact tab. Same user recovery policy (Goal/Loop or the recover-tabs switch; a compacting chat qualifies on its ticket's own evidence), one bounded reload under the shared per-conversation cooldown, retired by the revived page's first real activity. A stalled ordinary chat with nothing owed is deliberately left asleep.
- New `stalled` repair reason with its timeline label (`a suspended browser tab`); chat-scoped like `no-tab`/`silence`, so ordinary activity retires it.

## Test note

`test/desktop-input-maintenance.test.ts`'s `tabs.update` mock only clears `pendingUrl` when the patch navigates — matching Chrome, where a policy flip like `autoDiscardable` leaves a loading page's target intact.

## Validation

- New coverage: stalled/frozen reporting in the maintenance POST (`extension.test.ts`), revival reload of a discarded exact tab with no duplicate opening, discard protection of a freshly created input tab across passes; bridge-side handout for a stalled worker tab, the shared cooldown while Chrome keeps re-suspending, the ordinary-chat opt-out, the missing-open-tab guard, the compaction-source exemption, and malformed-report rejection.
- `test/bridge.test.ts`, `test/extension.test.ts`, `test/desktop-input-maintenance.test.ts`, `test/input-delivery-integration.test.ts`, `test/resume.test.ts`, `test/content-script.test.ts`: pass (1,451 tests).
- `npm run typecheck` passes. Full suite: the only reproducible failures on this machine are two pre-existing zh-CN Windows locale issues (`mcp.test.ts` PowerShell parser wording, `windows-keys.test.ts`), both confirmed on a clean `origin/main` worktree; the remaining full-run failures are load-induced flakes (temp-file rename EPERM, parallel-run timeouts) that pass when re-run in isolation.

Not live-verified against a real Chrome session — the exact discarded/frozen flag flow is covered at the harness level only.


totec448-spec:
Discarded/frozen tab recovery is incorporated through the existing maintenance owner. The integration uses fresh tab/document checks and elects an exact recovery target rather than reloading every matching tab. The original contribution is recorded in CONTRIBUTORS.md. Closing the already adapted proposal.

Reviewed head: `7bc90d9005a9`. Compared with published main `c5ab88714d3b` (2.1.14).

Issue #269, opened 2026-09-17
CoS companion debugging not working / continuos request timing out


jackbarunz:
### App version

2.1.13

### Operating system and version

Windows 11 Home

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT plan

Plus

### ChatGPT surface and workspace type

Chat; Personal

### What happened

Mid to long tasks unfortunately keep making the chat interrupting the connection, ending up in expiring. The CoS companion debugging doesn't seem to do the job.

<img width="809" height="977" alt="Image" src="https://github.com/user-attachments/assets/f313884c-b30f-4138-a7cf-66545b36e8c6" />

<img width="782" height="924" alt="Image" src="https://github.com/user-attachments/assets/ce82621e-7048-4d77-9f47-1e4941978db6" />

### Anything else

```text

```

Haz4rdovisk:
Something might have changed on OAIs end. Stopping the task and simply sending "continue" does the job for now.

totec448-spec:
So it only reloads like once per error message from chatgpt to not be sus.

BroNils:
> Something might have changed on OAIs end. Stopping the task and simply sending "continue" does the job for now.

Nope , did not work

<img width="1374" height="657" alt="Image" src="https://github.com/user-attachments/assets/c92623eb-fc03-4ada-9652-6624215da9de" />

jackbarunz:
> So it only reloads like once per error message from chatgpt to not be sus. 

Does this restriction come with the latest version? I don't remember this happening in 2.1.12

Haz4rdovisk:
> > So it only reloads like once per error message from chatgpt to not be sus.
> 
> Does this restriction come with the latest version? I don't remember this happening in 2.1.12

I don't think it's version related. Probabiy OAi rate limiting.. Have you noticed any timeframe pattern?  Or job related? Happens with any task?

Maximapple:
I have been chasing the same thing on 2.1.12/2.1.13 and have measurements that may save you the instrumentation. macOS here, Pro, so the numbers are from one install — but the mechanism looks provider-side rather than OS-specific, and the app-side consequences are not platform-specific at all.

## The timeout is not an independent failure

[...]

PR #274, opened 2026-09-17
Hand the brief back when ChatGPT refuses the resume send

Maximapple:
Measured on 2026-09-17: the brief reached the replacement chat, and ChatGPT
answered the send with its own "Message delivery timed out. Please try again."
No conversation id was ever assigned, so the page had nothing to name. It waited
out the forty-second lookup below, acked a send it could not bind, and the
journal gate stayed shut until the lease expired fifteen minutes later — the
brief gone, and the source chat left above the threshold that had filed the
ticket. The only visible trace was a replacement chat holding one unanswered
message.

That banner is not the ambiguity an armed click is protected from. It is ChatGPT
stating the message did not get through, and the destination half already has
the invariant that makes acting on it safe, in the words of
releaseContinuationSourceSendNow: "a chat that still has no id cannot have
accepted a message". So the existing `destinationLost` report is sent — the same
one the cleared-composer path has used since 2026-09-02 — and the brief is
offered to a fresh chat at once.

Narrow on both sides:

- Only the DOM layer's `recoverable` transport vocabulary counts. A rate-limit
  dialog and "Thinking failed" are not it and may not release an armed ticket.
- Only while this page is still on target and has no conversation id of its own.
  An id that does arrive outranks any banner beside it, and the loop takes it.
- A sighting must survive two consecutive polls, so a card that renders for one
  tick and is gone on the next is never ChatGPT's verdict on this send.

The app-side release is bounded, which the cleared-composer path never needed to
be: that one is a person pressing Escape, this one is automatic, and the failure
it reports is usually about ChatGPT rather than about this brief. Unbounded, a
bad afternoon would answer itself with a fresh chat every couple of minutes for
the six hours of AUTOMATIC_HANDOVER_TTL_MS. After three re-offers the brief stops
moving and the refusal is logged instead.

Tests: the reproduction, plus the two negatives (an id arriving alongside the
banner; a live-region notice outside the transport vocabulary) and the ceiling.
No protocol change — the wire shape and the app's handler both already existed.

---

**Why this is not what #145 was held for.** It does not release custody because acceptance signals are missing — delayed signals stay ambiguous and are still retained. It acts only on ChatGPT's own rendered transport failure *together with* the page still having no conversation id, which is the positive evidence the existing comment in `releaseContinuationSourceSendNow` already names: "a chat that still has no id cannot have accepted a message". A failure card that renders for one poll and is gone on the next is ignored, and an id that arrives outranks the card.

**Validation.** The reproduction fails against unmodified main and passes with the change, plus two negatives (an id arriving alongside the banner; a live-region notice outside the transport vocabulary) and the re-offer ceiling. `npm run typecheck` clean; bridge, extension and content-script suites green.

Root cause context and measurements: #269.


Maximapple:
Incorporated in  and closing accordingly. You retained the reproduction and solved it the better way — an unnamed resume keeps its dispatch custody instead of reporting a success it cannot name, so nothing is resent and the ceiling I wrote is unnecessary. Verified against current : the regression is there and the banner-driven resend is not. Thank you for the credit in CONTRIBUTORS.md.

PR #275, opened 2026-09-17
Compact the oversized chat whose turn ChatGPT just killed

Maximapple:
When a turn dies with a transport failure, the app asks the browser to reload
the chat. That is worth trying and often cannot work: after "Connection
interrupted" ChatGPT gives the page a fixed four minutes to produce the answer,
and a large chat does not come back inside it. Measured across 154 such
interruptions here, the ones that were rescued sat at a median of ~247k context
tokens and the ones that ended in "Message delivery timed out" at ~431k. Past
that size the reload cannot win the race.

The one remedy that changes anything is a smaller chat — and it was the one
thing that could not run. `considerAutomaticCompaction` requires
`chatIsWorking`, and a chat whose turn ChatGPT just killed has no turn, so the
call was never made and the chat stayed exactly as big as it was. One session
here ran that loop three times in a row on 2026-09-15: a 45-to-56-minute turn
died, the reload failed, nothing compacted, the next turn died the same way.

So a turn lost to a transport failure asks for compaction with `lostItsTurn`,
which is the only thing that waives the running-turn requirement. The
requirement itself stays for every other caller: it is the edge trigger that
keeps an old chat opened above the advisory line from compacting on sight. This
case is its opposite — the chat was working a second ago, it is over the line,
and the reason its turn died is that it is too big to recover.

Everything else still gates: the auto setting, `goalFencedChat`, an existing
continuation for the session, `autoCompactionReady`, a superseded conversation,
`automaticCompactionAllowed`, and the per-conversation filing lock. Both
`chatIsWorking` checks are waived together, since a chat called here has no turn
to re-read after the awaits either.

---

**Validation.** New regression plus the existing compaction suite; `npm run typecheck` clean; bridge suite green.

Root cause context and measurements: #269.


Maximapple:
Force-pushed one correction to the measurements this PR cites, before any review time goes into it.

The commit message and the code comment quoted token medians — ~247k for interruptions the reload rescued against ~431k for the ones that timed out. **Those are withdrawn.** I cannot reproduce them from the data I still have and cannot reconstruct the method that produced them, so they had no business reading like evidence in a source comment. The four-minute window is real but was quoted too precisely as well: 23 of 31 pairable `Connection interrupted` → `Message delivery timed out` sequences fall between 235 and 245 seconds, median 243.

What replaces them is stronger and does reproduce from the stored session events. Of 44 turns that saw `Connection interrupted`, the 23 that went on to time out had been running a median of **49 minutes** when it arrived; the 21 that did not had been running **2.6**. Duration is what decides whether the reload can win the race; chat size is only the proxy the app can act on before a turn starts.

No code change — the behaviour, the guards and the regression are identical. Only the comment and the message. Retracted in #269 at the same time.

Maximapple:
Incorporated in  and closing accordingly — current  carries the exact failed-turn guard, which is stricter than my  parameter because a historical banner can no longer earn a compaction. Verified against main before closing.

PR #276, opened 2026-09-17
Check once a minute that each ChatGPT page still has its recorder

Maximapple:
`restoreChatgptTab` can already repair a page whose injected world is gone, but
`restoreOpenChatgptTabs` only runs it when the service worker starts. That
sounds frequent and is not: the wake channel is a WebSocket, and an MV3 worker
holding one is not evicted, so a single worker lifetime routinely spans hours —
zero worker starts across 79 minutes of continuous work here. A world that dies
inside that window is never asked again.

The resulting state is the one the app cannot reason its way out of. The page
still renders, the model still calls tools through the MCP tunnel, so the work
looks alive — while no observation reaches the app at all. With no page evidence
there is no conversation the app can name, so an unattributed incident opens
with no repair candidate and nothing to reload, and the silence watch never sees
a chat because there is no grant and no open turn. Measured on 2026-09-14: the
recorder went quiet at 14:30, tool calls continued until 14:49, then 76 minutes
of standstill with not one line in the log, ended by reloading the tab by hand.

So the existing status poll also asks each ChatGPT tab, at most once a minute,
and hands every live one to `restoreChatgptTab`. Deliberately unconditional
rather than pinging first and skipping the tabs that answer: a healthy
content.js does not prove the independently running MAIN-world helper is still
there — `restoreChatgptTab` says exactly that in its own comment — and it pings
for itself, so re-running it is idempotent and costs one message per tab on the
healthy path. A page missing only fiber.js is not quiet but blind, and
`markedContinuationTurns` reads the resume marker out of Fiber turns, so a
replacement chat in that state can never reconcile the marker that commits its
own handoff.

Loading, discarded and non-ChatGPT tabs are skipped. Five scenarios cover
healthy, dead, stale-version, loading and discarded, plus the once-a-minute
bound.

Related: #167, where a replacement chat sits at `not-attempted` until the tab is
reloaded by hand. I cannot prove that is the same cause, but it is the same
signature from the app's side, and this closes the window in which nothing asks.

---

**Validation.** Five scenarios — healthy, dead, stale-version, loading, discarded — plus the once-a-minute bound. Extension suite green; `npm run typecheck` clean.

Related: #167. I cannot prove this is that issue's root cause, and the commit says so; what it does fix is that nothing asks for hours.


Maximapple:
Incorporated in  and closing accordingly —  is on current . Verified before closing.

PR #277, opened 2026-09-17
Assert usage amounts through the same formatter the renderer uses

Maximapple:
`npm run verify` fails on any machine whose locale is not en-US. Four
assertions in test/renderer-usage.test.ts compare rendered amounts against
literal en-US punctuation ('1.44', '0.48'), but the renderer formats money with
`Intl.NumberFormat(undefined, …)` on purpose — the amount is read by whoever
runs the app, in their own locale. On a de-DE machine the same correct render
reads '1,44 $', so these four fail while nothing is wrong:

  shows the calculated 256000 context cap and edits formula preferences …
  shows the calculated 400000 context cap and edits formula preferences …
  shows the Sol picker alias rate and preserves an explicitly cleared rate …
  combines equivalent recorded names in the table while keeping raw rate edits …

Reproduced on current main today. Asking the same formatter what the number
looks like here keeps each assertion on the amount, which is what it is
testing, instead of on the separator. Test-only; no production change.

---

Test-only, no production change, so sent straight as a PR rather than an issue first. Reproduced on current `main` today on a de-DE machine: the four assertions read `0,48 $` where they expect `0.48`.


Maximapple:
Incorporated in  and closing accordingly — current  asserts through , so the suite no longer fails outside en-US. Verified before closing. (Its remaining red check is the  flake, unrelated to this change; #288 addresses it.)

PR #278, opened 2026-09-17
Hold a handoff's tab until its command is done, not until ChatGPT names it

Maximapple:
A tab this worker opens for a command is held non-discardable while it still
carries its `clf=` marker and has no conversation of its own. For a resume that
is the wrong half of the window. ChatGPT names the conversation the instant it
accepts the brief, and the handoff is not finished there — the page still has to
reconcile that marker out of its own transcript before the app can commit. So
`opening` goes false, `nonDiscardable` does not contain a chat the app cannot
know yet, and the next maintenance pass actively hands the tab back to Chrome at
the one moment the page has to stay alive.

Measured four times on 2026-09-16/17. Chrome discarded the background tab and
content.js went with it. The turn kept running server-side and kept calling
tools through the app, so the work looked alive while no observation reached
anyone: one `session_start`, then dozens of tool calls and not a single turn. A
page that reports no turns can never reconcile the marker, so the handoff cannot
commit, so the session never moves — and the chat never becomes one the app
protects, which is what discarded the tab. Each half caused the other. The
replacement session's `origin` stays null throughout, which is exactly why it is
never in `managedConversations` and so never in `nonDiscardableConversations`.

A tab is now held while the command it was opened for is still live. Three
things end it: the app retiring that command, the app taking the conversation
over through `nonDiscardableConversations` (the handoff landed and needs no
help), and a thirty-minute deadline — longer than the destination lease, short
enough that a command which neither lands nor retires cannot pin a tab for
hours. The mapping is dropped when the tab closes and survives worker restarts
beside the existing protection state.

Also recorded for foreground tabs, which were not protected before: the person
can send one to the background a second later and the window is the same.

Validation: the regression fails against unmodified main — the tab is handed
back to Chrome — and passes with the change. Extension suite green. No protocol
change; `retiredCommands` already crosses the wire.

---

Root cause context and measurements: #269 (third item).


Maximapple:
Incorporated in  and closing accordingly — current  carries the command-ledger tab custody, which is better than mine: it validates against the app's live  rather than a retired list, pins the conversation id once learned, and guards a clock that moved backwards. Verified before closing.

Issue #271, opened 2026-09-17
Connection interrupted. Waiting for the complete answer


BroNils:
### App version

2.1.13

### Operating system and version

Windows 11

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT plan

Plus

### What happened

Problem occur when the task is a long running time consuming. At first i thought this was my internet problem, but i changed my internet and the problem still exists. Maybe Chat GPT patch this method now? but this is unaffected when its a short time task. 

# Webchat
<img width="846" height="677" alt="Image" src="https://github.com/user-attachments/assets/b7e3309c-99e4-4822-8411-ecc5ea126620" />

# Chat steroid

<img width="1500" height="858" alt="Image" src="https://github.com/user-attachments/assets/b98714db-7ba5-4189-8e70-12c3cb96fcf7" />

### Steps to reproduce

1. Just prompt for a long running time consuming task
2. This will happen

### Anything else

```text

```

BroNils:
This annoys me , because when i tried 'retry' it just redo all of the works has been done (re-read every file, everything is redo). Its like the context never stored after doing all that . Even tho the Chat Steroid thinks it has consumed about 500K context (the stats show)

BroNils:
# re-run

<img width="1354" height="210" alt="Image" src="https://github.com/user-attachments/assets/18a32214-df67-4962-b0b6-2fe6792d05a4" />

# first run (before error)

<img width="1363" height="583" alt="Image" src="https://github.com/user-attachments/assets/1796d69b-ef0d-4166-a65a-5e444bbccbf7" />

BroNils:
Aand i cant get the task done, because it is in a loop doing it all again whenever this interruption occur (click retry but it just redo all of it works like it forgot all context before). Chat Steroid thinks it 500K context but it did not match with agent behaviour

BroNils:
just realize, related #269

Maximapple:
Same signature here, measured on macOS across 154 of these, so the mechanism is not Windows-specific. Your instinct that it depends on task length is right, and it is the sharpest thing in the data:

[...]

PR #280, opened 2026-09-17
Recognise the page that stopped reporting, and reload it

Maximapple:
Follows the 2026-09-17 integration: this fixes a gate in `queueStalledTabRecovery` as it now stands, and adds the detector that reaches it.

## Root cause

Attribution is the health signal the app has, and on its own it is the wrong one — it proves the request-id join works, not that the page is still reporting. `noteCallAttribution` names one direction of that already ("a document can keep reporting turns and progress while its request-id reporting is dead"); this is the other, and nothing watched it.

A tool call happens only inside a turn. So an attributed call arriving while the app holds no turn for that chat is not a quiet page, it is a contradiction — and after three unbroken minutes of it, the page is the half that is wrong.

## Why the existing recovery could not help

`queueStalledTabRecovery` can reload exactly this page. It declined every time:

```ts
if (!working && !compacting && agent?.role !== 'worker' && …) return declined('no turn is running in it');
```

`working` reads the page's account of the turn, and `no turn is running in it` is precisely what a blind page always looks like. The gate asked the broken half whether it was broken — the same shape as `chatIsWorking` in `considerAutomaticCompaction`, one layer down.

## What this does

- Adds the detector: three unbroken minutes of attributed calls with no turn from that chat's just-filed summary. Any call that does carry a turn ends the stretch; the next one starts over rather than inheriting the minutes. A lagging browser poll cannot manufacture it.
- Lets `working` accept the app's own evidence — an attributed call inside the same window — so the reload is no longer refused for the one chat that needs it.
- Asks for the repair on the condition rather than behind the report, which is throttled to one a quarter hour across every chat. A reload that waits for a free report slot is a reload that does not happen.

Every other refusal that recovery had is unchanged: no session, tab recovery off, a slot that is not working, a chat that never called a tool, a browser action already pending.

## Measured

- A chat ran 06:45–08:51 with **195 attributed calls, one `session_start`, no `turn_start`**, no thought activity and no answer. Every turn-keyed guard was off: the silence watch had no turn to time, compaction read `chatIsWorking` as false, and when the turn died nothing noticed for 36 minutes.
- A later run in the same state reached **390k tokens across 453 calls** before its turn died where nothing could rescue it — no compaction for the same reason, and no handoff either, because a source send needs page evidence the page was not producing.
- Four chats ended that way in one day here, two of them handoff destinations (which #278 covers) and two of them the chat the person was sitting in, which nothing covered.

## Validation

Regression fails against unmodified `main` — neither the report nor the reload happens — and passes with the change. `bridge` 445, `extension` and `content-script` 825, typecheck clean, public-history privacy check clean. Two files, +111/-2.

Related: #269, #271 (the same incident from two other machines).


Maximapple:
Pushed two corrections to this branch, both found by running it on a real incident rather than by reading it. Worth reading before any review time goes in.

**The reload was not bounded.** I had relied on `queueStalledTabRecovery` declining while a browser action for the chat was already pending. That bound does not hold: a page in this state calls a tool every few seconds, nothing of the previous request survives to the next, and the result was **twelve reload requests in one minute** — thirty-five notices in the person's own timeline before I caught it. It is now one attempt per chat per stretch, and the counter clears the moment the page reports a turn again. My apologies for shipping that shape in the first version; the comment in the code now records that it was wrong and why.

**The app/extension boundary was silent, and that cost me an hour.** `queueBrowserRecovery` logs that a repair was asked for, and nothing said what became of it. When the chat stayed unrepaired, three possibilities were indistinguishable from the app side: queued and never handed out, handed out and refused by the browser, or carried out and ineffective. I resolved it with a unit probe of `performBrowserRepairs` driven from the real source — which showed the execution path is **fine**: a live tab with no `suspended` flag reloads exactly once, and the suspended contract still refuses a tab that is neither discarded nor frozen. So the remaining unknown is the handout, not the browser.

This branch therefore also adds the two lines that would have answered it in five seconds: the handout names which chats and reasons crossed, and the receipt says what the browser did with each, including the failure case which had no line at all. Both fire only when something actually happens — an idle pass hands over nothing and stays quiet. Happy to split that into its own PR if you would rather keep this one to a single concern.

**Field evidence from today, since it also validates the detector half.** On the run that prompted all of this: automatic compaction fired by itself at 406k, the brief was written and dispatched, the replacement chat began working — and its page reported one `session_start` and never a `turn_start`. Three minutes later this change recognised that and asked for the reload, where the previous code declined with `no turn is running in it`. The handoff still expired, which is why the boundary logging matters: I cannot yet tell you whether the reload was handed over at all.

`bridge` 445, typecheck clean, privacy check clean.

Maximapple:
**Correction to my previous comment, and it matters for review.** I wrote there that the unit probe showed `performBrowserRepairs` was fine and "the remaining unknown is the handout, not the browser". The probe was sound and **the question was wrong**: I drove it with a repair carrying no `suspended` flag, because I had grepped for that flag in `bridge.ts` and found none. The app does not set it — `background.js` derives it, `suspended: entry?.reason === 'stalled'`. So the case I probed is one that never occurs for this reason, and my conclusion was worth nothing.

**What actually happens**, now measured rather than reasoned, because the handout/receipt lines from the last push landed first:

```
19:11:28.806  …6aac1e2c… has been calling tools for 4 minutes with no turn reported by its page
19:11:28.807  …is a stalled browser tab — asking the browser to reload the exact chat once
19:11:28.818  handing the browser 1 repair(s) — 6aac1e2c:stalled
              (no receipt, ever)
19:14:50      the same, again
```

Against, for the same chat, the same session, minutes apart:

```
19:06:30.633  handing the browser 1 repair(s) — 6aac01be:compaction
19:06:30.667  the browser reloaded the compaction repair          ← 34 ms
16:41:10.007  handing the browser 1 repair(s) — 6aac01be:silence
16:41:10.114  the browser reloaded the silence repair            ← 107 ms
```

So `silence` and `compaction` are carried out; `stalled` is handed over and never executed. The cause is the contract, working exactly as written:

```js
const candidates = live.filter((tab) => conversationForTab(tab) === conversationId &&
  (!suspended || tab.discarded === true || tab.frozen === true));
```

"Suspension grants a reload of a still-suspended shell, never a new tab" is right for the shells `stalledConversations` reports. A page that is alive and merely blind is neither discarded nor frozen, so it is filtered out, `target` is undefined, and the pass continues silently. **My detector was routing a live page through a contract written for dead ones.** That is my bug, not a defect in your recovery.

The third commit fixes it by giving the blind page its own reason. Everything else in `queueStalledTabRecovery` is shared and unchanged, because every refusal it has is still wanted — no session, tab recovery off, a slot that is not working, a chat that never called a tool, an action already pending. Only the shell contract is not, and `suspended` stays false for the new reason without the worker needing to know the word, since it derives the flag from `stalled` alone. New test at both ends: the app takes the new route and not the shell one, and a live tab reloads for `blind` and does not for `stalled`.

`bridge` + `extension` + `content-script`: **1,272 passed**, typecheck and privacy clean.

Worth saying plainly: this whole thing was invisible for a day and became obvious within minutes of the boundary logging existing. Six Compact & Resume handoffs failed here today, and the request→no-receipt asymmetry is what named the cause.

Maximapple:
Pushed `d1bcbc3` — a correction to this PR's own give-up line, prompted by watching it fire.

The line said the page "needs the extension reloaded". It fired today at 12:42:55 UTC on a chat whose extension was demonstrably healthy: reloaded four hours earlier, reading the page model, and acknowledging every repair handed to it within milliseconds — three of them in the preceding ten minutes, each executed in 13-32 ms.

What had actually happened was the other cause. ChatGPT ended the turn on its own side (`Message delivery timed out. Please try again.`, three times, `lastTurnOutcome: failed`) while the connector kept answering tool calls. The work was alive and arriving; there was simply no turn on the page to attribute it to, and no fresh document could find one. So the message named a remedy that cannot work, at the exact moment the app has stopped trying anything else — the worst moment to send someone after the wrong thing.

The full sequence from the log, for the record:

```
12:35:37  has been calling tools for 3 minutes with no turn reported by its page
12:35:38  the browser reloaded the blind repair                      (13 ms)
12:39:08  the browser reloaded the blind repair                      (2nd)
12:42:55  did not report a turn after 3 reloads — not reloading it again
12:42:55  the browser reloaded the blind repair                      (3rd)
```

The bound behaved exactly as designed — three attempts on the condition, the notice throttled separately, then it stops. Only the sentence was wrong.

It now reports the observation and leaves the diagnosis open: the calls are still arriving, so the work is alive; what is missing is a turn to attribute them to. A stale extension does that and is worth ruling out first. So does ChatGPT ending the turn while the connector keeps answering — and that one no reload can repair, because there is no turn left for a fresh page to find.

Nothing about the bound or the routing moves. The test now pins the three claims the sentence makes instead of its old wording.

Full suite on the branch: **4906 passed**, typecheck and privacy clean.

Maximapple:
Pushed `4c34d07`. This one came from a user asking the obvious question I had not: *why does it need so many attempts?*

## What the field says

Two blind stretches on one machine today, seven reloads between them:

```
12:35:37  blind for 3 min → reload
12:39:08                  → reload
12:42:55  gave up after 3 → reload
12:51:50  still blind, now 19 minutes
15:09:30  blind for 3 min → reload
15:31:04                  → reload
15:34:18                  → reload
15:37:26  gave up after 3
```

**Not one of the seven restored turn reporting.** Every one of them was on a chat whose `lastTurnOutcome` was already `failed` with no turn since — ChatGPT had ended the turn on its own side while the connector kept answering tool calls.

And the give-up line, since `d1bcbc3`, says exactly that: *a reload cannot repair it, because there is no turn left for a fresh page to find*. The code was saying it three reloads later than it knew it.

I should also correct myself on the cost. I first read these numbers as the repair damaging a working page, and that is wrong: work continued within a minute after all seven reloads, measured the same way. What they actually do is interrupt a working page three times for an outcome that cannot follow, and bury the remedy that does work under three failed attempts.

## The change

The reload is skipped when `lastTurnOutcome` is already `failed`, and the notice says which case it is — ChatGPT ended the turn, no reload undoes that, and sending anything in the chat gives it a turn again.

The repair keeps everything it was built for. A page whose *reporting* broke while ChatGPT is fine has no failed turn behind it, still gets its three attempts, and a fresh document does restore it there. That is the case this PR was written for; the failed-turn case was collateral that the detection happened to catch too.

The regression asserts on the request line rather than on `/status`, because a queued repair can be handed out and receipted between two polls — my first version asserted on `/status`, passed without the change, and proved nothing.

It also sits at the end of the file, with the reason written down: the blind-work notice is throttled once a quarter hour **across every chat**, so any test driving a blind stretch consumes that slot for whatever runs next. Placed earlier it silenced the notice `says which chat is calling tools…` asserts on.

Full suite on the branch: **4907 passed**, typecheck and privacy clean.

totec448-spec:
The useful recovery handout/result diagnostics were incorporated and credited. The additional blind-page reload detector remains excluded: separated observations do not prove continuous blindness, and its attempt count advances before a browser action is confirmed. The demonstrated oversized-chat gap is addressed by the already incorporated #295 adaptation.

Reviewed head: `4c34d07da3a7`. Compared with published main `c5ab88714d3b` (2.1.14).

Issue #283, opened 2026-09-17
Fiber-derived observations have collapsed: turns report, page_tool does not, and handoffs cannot reconcile their marker


Maximapple:
### App version

2.1.13 (and the current `main` snapshot `1ede459`)

### Operating system

macOS 15 (arm64)

### Was the Chrome extension connected?

Yes, and its version matches the app

### What happened

Fiber-derived observations have collapsed here over four days while content.js keeps working normally. I have no root cause, so this is a measurement report rather than a diagnosis — but the consequence is severe enough to be worth checking on other installs, because `markedContinuationTurns()` reads the resume marker out of Fiber turns. No Fiber data means a replacement chat can never reconcile the marker that commits its own handoff.

**Counted from the local session journals, `kind: "page_tool"` per `turn_start`:**

| Day | Sessions | Turns | Tool calls | `page_tool` | per turn | Sessions with any |
| --- | --- | --- | --- | --- | --- | --- |
| 09-14 | 28 | 99 | 7,580 | 521 | **5.3** | 22 of 28 |
| 09-15 | 25 | 51 | 4,408 | 60 | 1.2 | 13 of 25 |
| 09-16 | 22 | 28 | 2,839 | 67 | 2.4 | 6 of 22 |
| 09-17 | 14 | 13 | 1,253 | **1** | **0.08** | 1 of 14 |

A factor of 66 per turn, and today one event across 1,253 tool calls.

**The two halves fail apart.** One chat today: `session_start`, two `turn_start`, one `turn_end`, 272 tool calls, and **zero** `page_tool`. So content.js is observing the DOM and reporting turn boundaries, while nothing Fiber-derived arrives at all. Yesterday's comparison chat: 23 tool calls, 16 `page_tool`, same shape otherwise.

**And the data is on screen.** The reasoning block is visible in the page the whole time — the person watching it described the collapsed "Stopped thinking" state while the app received no `page_tool` for that chat. So the thought messages are rendered and are not reaching the app.

**Consequence I can demonstrate.** Five Compact & Resume handoffs here today ended `dispatched-unresolved` with `to: null`: the brief demonstrably landed (the replacement chat was on its 49th, 180th, 272nd tool call), the chat worked, and the ticket expired anyway. That is exactly the failure mode the comment in content.js warns about — "the brief was sent, the prime in chat B read its predecessor session and went to work, and the marker was never redeemed".

### What I ruled out

[...]

PR #284, opened 2026-09-17
Report the page-model helper only the page can see

Maximapple:
Closes the observability half of #283. This does not fix the fault — I still cannot say what causes it — but it separates the two candidates, which is the thing nobody can currently do from the app side.

## Why it matters

Everything Fiber-derived reaches the app through the MAIN-world helper: thought activity, the canonical assistant message keyed by ChatGPT's own id, and the `[[CLF-RESUME]]` marker that `markedContinuationTurns()` reads to commit a handoff. When it stops delivering, the chat works normally and its handoff can never commit — and from the app that looks exactly like a page that is merely quiet.

Measured on one install across four days, `page_tool` per `turn_start`: **5.3 → 1.2 → 2.4 → 0.08**, the last being one event across 1,253 tool calls, while `turn_start` and `turn_end` arrived throughout. Five handoffs that day ended `dispatched-unresolved` with the brief demonstrably in a chat that was on its 272nd tool call, and the reasoning block was rendering in the page the whole time. Nothing in the log said why.

## What it adds

The page already decides this and had no way to say it — `fiberPresent` is settled only after a repair attempt. So the activity pull it already makes carries one word:

- `absent` — the helper did not answer even after a repair. An injection that is not landing, which the app can act on.
- `empty` — it answered and had no turns. A change in what it reads, which cannot be fixed from here.
- `ok`.

The worker forwards it verbatim and drops anything else at the join; the app logs the two faults and the recovery, once per chat per state change.

## Shape

`absent`/`empty`/`ok` on `/activity`, additively — an older app ignores an unknown query parameter and a newer one treats its absence as unknown, so no protocol change. This is the same shape as `stalledConversations`.

## Validation

A test per link, because this field crosses the three files where two checkpoint fields have already been lost with every unit test still green — that join is what `scripts/verify-compact-chain.mjs` exists for:

- content.js sends it (and says `absent` when no helper answers).
- background.js forwards exactly the three accepted words and drops `maybe` and absence.
- bridge.ts distinguishes the two faults, reports each once per state change, and says so again on recovery.

`bridge` + `extension` + `content-script`: **1,272 passed**. Typecheck clean, public-history privacy check clean. One word on the wire, three files, no behaviour change.


Maximapple:
The red Windows check here was not this change — it was the accessibility probe timing out on a loaded runner, which I have since traced and fixed separately in #288.

I cannot re-run a job in this repository, so I have appended that one commit (`Give the Windows accessibility probe room a loaded runner needs`) to this branch to get a deterministic Windows run for the change under review. It is the same commit as #288; if #288 lands first it disappears from this branch on rebase, and if you would rather review this one without it, say so and I will take it back out.

Maximapple:
Correction to this PR's own evidence, pushed as `ad49222`.

The comment and my commit message said `page_tool` observations per `turn_start` fell to **0.08** on the last of four days — "one event across 1,253 tool calls". That number is wrong. I recounted it over the full session journals rather than the partial day I had originally measured mid-afternoon:

| day | `page_tool` | `turn_start` | per turn |
|---|---|---|---|
| 2026-09-14 | 521 | 99 | **5.26** |
| 2026-09-15 | 60 | 51 | **1.18** |
| 2026-09-16 | 67 | 28 | **2.39** |
| 2026-09-17 | 21 | 16 | **1.31** |

So there was no collapse to nothing, and I withdraw that claim and the factor that went with it. I had already retracted it on #283; it should not have stayed in this branch, and that is my mistake.

What is left still argues for the change, and argues for it more honestly: the yield steps down about fourfold after the first day and stays down, while `turn_start` and `turn_end` arrive throughout — the isolated world is alive the whole time. And the fault this PR reports is a per-page one that no daily aggregate can show: in that same window five handoffs ended `dispatched-unresolved` with the brief demonstrably in a chat that was on its 272nd tool call. A per-chat health line is exactly what distinguishes that page from a quiet one; a day's average never will.

Checks are green on the corrected branch.

Maximapple:
I have found the actual cause of the handoffs I cited here, and it is not this. Retracting the diagnosis in this PR's comment before it misleads anyone, including me.

**What I claimed:** that the MAIN-world helper had stopped delivering, and that five handoffs ended `dispatched-unresolved` because of it.

**What the data says:** the helper was alive and reading the page model in every one of those sessions. `page_tool` counts ChatGPT's *rendered activity rows*, which vary with what ChatGPT does in a turn — it is not a health signal, and I used it as one. The signal that does prove the helper is reading is request-id attribution, since `request_id` requires `message.metadata.request_id` out of the page model, and it is **100% across every session in that window**, including every session I had pointed at as evidence of a dead helper. Two sessions started in the same minute, one with 16 `page_tool` per turn and one with none, both attributing every call by request id.

**The real cause is #291:** ChatGPT's composer began round-tripping inserted text through its own Markdown serializer, which escapes ASCII punctuation, so `[[CLF-RESUME:<token>]]` arrives as `[[CLF-RESUME\:<token>]]`. Every escaped marker failed to commit and every clean one committed, with no exceptions across the full recorded set.

That leaves this PR without the evidence it was argued from. The feature may still be worth having — a per-chat line saying whether the page-model helper answers is a thing nothing reports today, and I would have found #291 faster with it — but it is now a diagnostic looking for a justification rather than a fix for a measured fault, and that is a weaker case than the one I made. Your call: I am happy to close it, or to rewrite the comment so it claims only what is supported. Say which and I will do it.

Sorry for the noise. The measurement was real; the inference I drew from it was not.

Maximapple:
Pushed `70237cf` — I shipped this reporter locally, watched it run for four hours, and it warns about healthy pages.

Every reload tears the MAIN-world helper out of the page and the worker puts it back, so an ordinary reload produces `absent`, often `empty` while the transcript is still mounting, and then `ok`. Reporting the first sighting means reporting that sequence.

Measured today across five reloads on one machine — every state change this reporter emitted, with the recovery time:

| chat | degraded | recovered after |
|---|---|---|
| 6aac54fe | absent | 1.96 s |
| 6aac5615 | absent | 9.40 s |
| 6aac54fe | absent | 1.50 s |
| 6aacf6ae | absent | 1.48 s |
| 6aacf6ae | absent → **empty** | 2.11 s / **0.04 s** |

**Every single one healed itself.** The `empty` lasted 44 milliseconds: the log announced that a chat "contributes no thought activity, no authored answer and no resume marker" about a page that was reading again before the sentence finished. Two warnings per ordinary reload, and the one that reads most alarming is the one that was least true.

So a degraded state now has to last fifteen seconds before it is worth a line — past the slowest recovery above by some margin — and the recovery notice only follows a warning that was actually given. A state change inside the window restarts the stretch rather than satisfying it, and a clock that moves backwards restarts it too.

What survives that is the state this PR was built for: a helper that is **not** coming back by itself, which is the one a person can act on. Nothing else about the change moves.

The test now drives the clock through all three cases — the blip that says nothing, the fault that lasts and is reported once, and the recovery that follows only a real warning.

I would rather find this in my own PR than have you find it in your log. `test/bridge.test.ts` 445 passed, full suite 4904 passed, typecheck and privacy clean.

totec448-spec:
The latest reporting-gap correction is now adapted as well as the earlier helper-health work. A 90-second gap or backward clock starts a fresh 15-second observation grace; current hidden-page polling still counts as continuous observation. The original regression failed before the change, and the boundary, rollback and existing health tests now pass. The feature remains diagnostic only. Published in #316 (https://github.com/totec448-spec/chat-on-steroids/pull/316), merge commit `962d040b423f`. This is a source integration for the next release; it does not install or publish a new application version.

Reviewed head: `32d0b6fcdebe`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #286, opened 2026-09-17
Stand down from a recovery reload that would discard a draft

Maximapple:
Found while auditing my own #280 for safety, and it turns out to apply to `main` as it stands, independently of that PR.

## The gap

`performBrowserRepairs` ends in a bare `chrome.tabs.reload(target.id)`. A reload discards whatever the person has typed and not sent, and nothing on that path asks whether there is anything to lose.

The contrast inside this same file is what convinced me it was unintended rather than a decision: every close path proves the composer is empty first, through `clf-tab-close-check`, and the comments there are emphatic about it — "the draft was preserved and nothing was sent". The reload paths ask nothing at all.

It is not hypothetical. `silence`, `compaction` and `no-tab` all target a page that is generating, and generating is exactly when someone is writing their follow-up.

## Why not `clf-tab-close-check`

It also requires `!generating`, and the page a recovery reload targets is generating by definition. Using it would refuse every reload this recovery exists for.

So `clf-page-status` — which the worker already sends on other paths — carries one more boolean, and the reload stands down when the page says there is a draft.

## Standing down is reported, not silent

It reports the handout as failed rather than skipping quietly, so the app can show it and the repair stays retryable once the draft is gone. A reload that silently does not happen is how a chat ends up unwatched with nobody knowing why — I spent an afternoon on exactly that shape this week.

## What it does not change

Only an answer counts. A page whose content script is gone cannot answer and cannot be holding anything a reload could preserve, so that silence reloads exactly as before — which is also the state most of these recoveries exist for. Three cases, three tests: a reported draft (no reload, one failure receipt), an empty composer (reload, one success receipt), and no answer (reload).

`extension` + `content-script`: **830 passed**, typecheck and privacy clean. Two files plus tests.

The three isolated contexts that drive `performBrowserRepairs` now supply the probe, because they model the environment it runs in — and that they did not is what surfaced the new dependency immediately.


totec448-spec:
Draft protection is already incorporated for ordinary and compaction recovery. Current code rechecks the exact document and browser claim and preserves attachments as well as text; an unanswered probe is not treated as permission to discard a draft. The adapted contribution remains credited in CONTRIBUTORS.md.

Reviewed head: `e44f9f16ca62`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #288, opened 2026-09-17
Give the Windows accessibility probe room a loaded runner needs

Maximapple:
`Verify Windows x64` is flaking, and it is blocking review of unrelated work.

## What fails

```
AssertionError: WINDOWS_ACCESSIBILITY_PROBE_OK: expected true to be false
 ❯ test/computer-windows-accessibility.test.ts:211:43
```

Line 211 is `expect(timedOut, …).toBe(false)`, so the failure is the sixty-second bound, not the probe's output — the message in the assertion is just the captured stdout, which reads confusingly like an assertion about the probe itself.

## Why it is a flake and not a signal

Three hosted runs failed it on 2026-09-17 while the same job passed on other runs of the same code. One of the failures was on a PR that changes nothing but number punctuation in `test/renderer-usage.test.ts` — there is no path from that to Windows UI Automation. A contributor independently reported "Windows UI Automation" among failures that reproduce on clean `main`, in #269.

## The change

The bound exists so a hung probe fails instead of hanging for ever, which is worth keeping — but at sixty seconds it was asserting how fast a loaded runner starts PowerShell, builds a WPF window and drives several UI Automation round-trips through it. That is not what the test is about.

150 seconds still catches a genuinely stuck probe and stops catching a busy one; the per-test budget moves with it because it has to outlast the bound it contains. Test-only, one file, no production change.

If you would rather keep sixty and make the runner faster, or split the fixture build out of the timed section, I am happy to do either instead — the point is that this check currently fails for reasons unrelated to the change under review.


totec448-spec:
The accessibility probe already has the proposed 150-second process allowance and 180-second enclosing test budget in current main. Closing the incorporated test change; this timing allowance does not assert that a native runtime test was executed during this cleanup. Credit is retained in CONTRIBUTORS.md.

Reviewed head: `be3fbbc01318`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #289, opened 2026-09-18
Say a refused recovery once a minute, not once a status poll

Maximapple:
Found in my own log while watching a long run: at 03:05 today the same sentence about the same chat appeared twice inside a minute, and it does not stop while the condition holds.

## What repeats

`queueBrowserRecovery` and `queueStalledTabRecovery` both log their refusal from inside a `declined()` closure, and both are reached again on every status poll. A conversation that cannot be repaired right now — no tab to reload, or a suspended shell the contract will not grant a second one for — is therefore refused on every pass, with a line each time.

The refusals themselves are right. It is the cadence that is wrong: the condition changes on the scale of minutes, the poll runs on the scale of seconds, and the log is what the user reads when they want to know what the app is doing.

The visible cost is that the timeline scrolls. I lost 35 lines of a user-facing recovery timeline to one repeating refusal this week, which is how a second, unrelated problem in the same window went unnoticed.

## The bound

A helper keyed by `<site>:<conversation>:<cause>` keeps the first refusal and drops repeats for a minute. Consequences:

- a different chat speaks up immediately;
- the same chat with a different cause speaks up immediately;
- the same refusal an hour later speaks up again, because the minute has passed and it is news once more.

The key deliberately includes the cause, so a chat that stops being "no tab" and starts being "suspended" is not silenced by the earlier line.

The map is bounded (500 entries, oldest 100 dropped) so a long session cannot grow it without limit, and a clock that goes backwards falls through to logging rather than to silence — the same shape the other timers in this file use.

## What it does not change

Nothing about whether a repair is granted, attempted or retried. Only how often the refusal is narrated. `logWarn` paths for genuine failures are untouched.

Regression test drives the real refusal twice in a row and asserts one line, then advances the clock a minute and asserts the second — `test/bridge.test.ts`: **445 passed**, typecheck and privacy clean. One production file, one test file.


totec448-spec:
Current bridge.ts already bounds refusal notices by conversation and cause for one minute, including backward-clock handling and bounded map retirement. The recovery decision itself remains unchanged. Closing the incorporated patch with attribution retained.

Reviewed head: `b012cfd0b20e`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #291, opened 2026-09-18
Read the continuation marker ChatGPT now escapes as Markdown

Maximapple:
Compact & Resume stopped landing here, and this is the reason. It is not in this app's logic; it is one transformation ChatGPT's composer started applying to the text this app sends.

## What the transcripts say

The composer is a rich-text editor, and text inserted into it is round-tripped through ChatGPT's own Markdown serializer before it is sent. That serializer began escaping ASCII punctuation. The brief goes in as `[[CLF-RESUME:<token>]]` and arrives in the transcript as `[[CLF-RESUME\:<token>]]` — and for a token that starts with an underscore, as `[[CLF-RESUME:\_<token>]]`. One brief carried both.

Every resume marker in one install's session store, oldest first, against the outcome of that handoff:

| delivered | marker as rendered | outcome |
|---|---|---|
| 09-03 … 09-14 (56 handoffs) | `[[CLF-RESUME:<token>]]` | committed |
| 09-11 10:17 | `[[CLF-RESUME\:R3--bW\_9…]]` | failed |
| 09-15 05:45 | `[[CLF-RESUME:IAeHQRnF…]]` | committed |
| 09-15 08:48 | `[[CLF-RESUME:td0K-U1t…]]` | committed |
| 09-15 16:55 | `[[CLF-RESUME\:USm9G\_tI…]]` | failed |
| 09-16 14:27 | `[[CLF-RESUME:89bC0KSY…]]` | committed |
| 09-16 18:10 | `[[CLF-RESUME\:iNHBs\_C0…]]` | failed |
| 09-17 04:45 | `[[CLF-RESUME\:nWyfk2Qm…]]` | failed |
| 09-17 07:59 | `[[CLF-RESUME\:S77o6s9J…]]` | failed |
| 09-17 15:05 | `[[CLF-RESUME\:ZIwLOydX…]]` | failed |
| 09-17 19:19 | `[[CLF-RESUME:\_dv0eLoS…]]` | failed |

Every escaped marker failed. Every clean marker committed. No exceptions in the recorded set. The first escaped delivery is 09-11 and it becomes the normal case from 09-15 on, which is what a gradual rollout looks like from one machine — and why this felt like a slow decline rather than a break.

## Why it costs the whole handoff, twice

Both halves of the commit compare exact text, so both fail on the same rendering:

- `markedContinuationTurns()` matches `CONTINUATION_MARKER` against the page's text. `\:` does not match, and `\_` is not in the token class, so the marker is never reconciled.
- the send receipt (`matchesSubmittedUser` via `CLF_DOM.send`) compares the rendered user message with what was typed. It differs by the escapes, so no receipt, so no ack — and without the ack the app never learns which conversation the replacement chat became.

That is the `dispatched-unresolved` state, from both directions at once: twelve continuations here between 09-14 and 09-17 ended that way, each one abandoned at its six-hour TTL, each with its brief sitting in a chat that was working normally.

## The change

One normaliser, applied only to text read back out of the page, and only after an exact match has already failed:

- `markedAs()` retries `CONTINUATION_MARKER` against an unescaped copy of the marker's first 200 characters — never the brief behind it.
- `matchesSubmittedBootstrap()` accepts a rendering that differs from the sent brief only by escapes, and is used **only** for the bootstrap send and its receipt.
- `resumeBootstrapMatches()` gets the same second chance, in the same shape as the NBSP and CRLF artifacts already canonicalised there.

Nothing that is about to be sent is ever unescaped, and an exact match is always tried first, so a page that stops escaping needs no further change.

**`matchesSubmittedUser` itself is deliberately untouched.** My first attempt did relax it, and `retains a reused-document send boundary when canonical Fiber text arrives before activity identity` caught that immediately: that comparison also answers "is this the message the app authorized me to send", where exact text is doing real safety work — an input the app refused to acknowledge must not acquire a send boundary because its text looks right. The bootstrap is the narrower question, asked about a fresh document with no other user message and a 16-64 character random token in the text.

## Tests

Three regressions, using the exact bytes from the transcripts above: the marker commits when escaped (fails without the change — verified by reverting the one line), the destination acks when the rendering is escaped, and the recorded bootstrap is still accepted as provenance. Plus a negative: an escaped marker on a brief whose *words* changed is still refused.

`content-script` 648, `bridge`, `extension`, `goal-resume-handoff`: all green, typecheck and privacy clean. Two production files.


Maximapple:
Second commit pushed — the page half was not the whole of it.

Auditing for the same shape elsewhere turned up **four more readers of the shared `CONTINUATION_MARKER`**, all of them reading text that came back out of ChatGPT, so all of them broke on the same rendering:

| where | what stops working |
|---|---|
| `src/main/bridge.ts` | the bootstrap boundary of a resumed chat is never found |
| `src/main/session/store.ts` | the newest RESUME message is not located, so `resumeBoundary` stays 0 |
| `src/renderer/chat.ts` (fold) | a Compact & Resume is not folded into its card at all |
| `src/renderer/chat.ts` (strip) | `[[CLF-RESUME\:<token>]]` is left in the transcript in full |

The renderer pair is the user-visible half, and it is what someone would actually report: three unfolded rows with a raw marker sitting on top of them, on every resumed chat since the escaping began.

`continuationMarkerOf()` in `src/shared/session.ts` now answers for all four. It tries exact text first and the escaped form second, and returns the text that matched so the strip removes whichever form was recorded rather than half of it. A token is 16-64 characters of `[A-Za-z0-9_-]`, so a form carrying a backslash before one of them cannot be anything but this marker escaped — the length requirement is what makes reading it that way safe.

The new renderer regression drives the whole fold from an escaped recording; without the change it builds zero cards instead of one (verified by reverting the one line).

Full suite on this branch: **4988 passed**, typecheck, licence notices and privacy clean.

Maximapple:
Third commit — one more reader, found by asking which other prompts this app types into ChatGPT itself.

The **worker bootstrap** is one. It goes into its fresh chat through the same insertion path as a resume brief, so the composer escapes it the same way, and the comparison in `readActivityEvents`' caller that proves the opening row *is* that bootstrap is exact text. It stopped matching, `bootstrapMessageId` stayed null, and with it went the boundary that keeps a predecessor chat's repair notices out of the new chat's feed — the exact symptom the comment above that code was written for.

Worth recording how the first attempt at it was wrong: I unescaped the *result* of `userPromptText(...)`. The escaping lands on the prompt's own framing too, so extracting first and unescaping afterwards compares the wrong slice, and the new regression failed on it. Undoing it before the prompt is parsed is what works.

`unescapeMarkdown` also moves to `shared/session.ts`, beside the marker it belongs with, so the bridge, the resume provenance check in `session/handoff.ts` and the renderer share one definition rather than keeping private copies that can drift apart.

That is every reader I can find of text this app typed into ChatGPT: two in `extension/content.js` (marker and send receipt), four in the app (`bridge.ts` boundary, `session/store.ts` RESUME lookup, renderer fold and renderer strip), and this one.

Full suite on the branch: **4988 passed**. One earlier run on this branch had a single unrelated failure and the re-run was clean — this branch does not carry the flake fix from #292, which is where that belongs.

totec448-spec:
All three useful additions are represented in current main: punctuation-escaped resume marker readback, shared app/renderer marker handling, and worker bootstrap matching. Literal brief text and ordinary authored-message receipt checks remain strict. The contribution and integration boundaries are documented in CONTRIBUTORS.md.

Reviewed head: `07d58430e1b8`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #292, opened 2026-09-18
Ask again for the unread bound, as its own refusal instructs

Maximapple:
`refuses new commands at the unread-result bound, delivers a result, then admits after automatic receipt` fails on roughly one full-suite run in three. I measured ten consecutive full runs on macOS: **3 red, 7 green, the same single test every time**, with all other 5081 tests green in all ten. Run alone — even alone under load, three times — it passes every time, which is what kept it looking like a product defect rather than a test one.

## It is not the product

`acknowledgeCompletedOutput` retires an offered page only for a call that started strictly after that page's publication completed:

```ts
if (!offered || offered.publication.failed || offered.publication.completedAt === null ||
    startedAt <= offered.publication.completedAt) continue;
```

That rule is right: a page counts as read once it demonstrably reached the caller, and an older concurrent response must not advance a newer delivery cursor.

The test issued its follow-up `exec_command` in the same instant as the refusal it was reacting to. On a loaded runner the publication tail has not landed by then, so nothing is retired and the unchanged count refuses the call a second time. The captured failure says so directly — the second call reports all four results still waiting, and delivers one page of its own:

```
EXEC_RESULTS_UNREAD: 4 completed background results are still waiting for this session.
… Explicit write_stdin is also available for IDs 3119, 32698, 79596, 85799.
--- Background command results ---
Background session 32698 completed … owed-3
```

No production caller can hit this: the model cannot issue its next call before it has received the response carrying the page, which is exactly the condition the rule asks for.

## The change

One test, no production code. The follow-up now does what the refusal text tells a caller to do — read the delivered output, then ask again. Each refusal retires one more session, so the bound falls within an attempt or two; the ceiling of eight is headroom for a loaded runner, not a second mechanism, and each refusal is still asserted to be the `EXEC_RESULTS_UNREAD` one.

A real break still fails: a bound that never lifts exhausts the loop and the final assertion reports the last refusal in full.

`test/mcp.test.ts` green, typecheck and privacy clean.


totec448-spec:
The unread-result regression was adapted to await actual completed-output publication before making the next invocation. It therefore covers the publication ordering problem without repeatedly attempting new commands. Closing this already incorporated test correction with credit retained.

Reviewed head: `8e9db4f3e7a9`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #294, opened 2026-09-18
Reduce Sahur pet CPU use without changing animations

totec448-spec:
Enabling the Sahur pet kept a permanent animation-frame loop running at the display refresh rate, even while its sprite was unchanged or its context menu paused the animation. On a 200 Hz display this meant roughly 200 DOM projections per second.

The existing animation owner now supplies the next frame, phase or autonomous-action deadline. Stationary frames sleep until that deadline; walking, carrying and throwing retain continuous interpolation. Pausing, hiding and disposal cancel pending work, and the renderer writes only changed DOM values. All 96 frames, artwork, timings, scale and choreography remain unchanged.

In an isolated Electron fixture on the same machine, idle CPU changed from 0.841% to 0.043% and normal autonomous behavior from 0.778% to 0.107%. These are controlled fixture measurements, not a guarantee of whole-application CPU usage. The reproducible harness records CPU deltas, renderer work and animation callbacks, and flags invalid process-counter samples.

Validation: 28 pet tests and TypeScript checks passed; the complete working-tree verification passed 5,244 tests with 45 skips. Production build, full Electron interaction acceptance and a fresh restart check passed, including both actions, dragging, interruption, themes, zoom, reduced motion and restored position. The idle/menu regressions were observed failing on the original implementation. GitHub checks validate this focused public branch separately.

The separate live-plugin CI step now runs its five tests with one worker. Two Windows runs passed ordinary verification but hit cold Python-server startup timeouts and locked-directory cleanup while multiple package-install suites ran concurrently. All five live tests pass locally in sequence. This CI scheduling change preserves every assertion and all production timeouts; the final GitHub run checks it on each platform.


redzrush101:
Yeah this seems important

PR #295, opened 2026-09-18
Compact the oversized chat that is working with no turn on its page

Maximapple:
**Stacked on #280** — it uses the blind-work detection that PR adds, so please take #280 first; this branch contains it and adds one commit on top.

## The gap

Automatic compaction has two triggers and both ask the page: one runs on page evidence of a live turn, the other on a turn that just failed. `chatIsWorking()` is `generating || activeTurnId`. A chat whose page has lost its turn while the connector keeps answering satisfies neither.

I watched that state for half an hour today:

```
12:32       ChatGPT ends the turn (three "Message delivery timed out", lastTurnOutcome: failed)
12:32-13:03 500 attributed tool calls keep arriving; files edited throughout
            context 311k → 356k, about 1,850 tokens a minute, threshold ~400k
            activeTurnId: None the entire time
12:51       "has been calling tools for 19 minutes with no turn reported by its page"
```

No live turn, and no *new* failing turn — the failure had happened while the chat was still under the threshold. So nothing could file a ticket, and the chat was heading past the cap with the only remedy being the person noticing and typing into the chat by hand. Which is what I told them to do, because the app could not.

## The change

The app already knew. `noticeBlindWork` had been logging this for nineteen minutes; that knowledge just never reached this decision. A confirmed blind stretch now offers it, on the same condition as the repair rather than on the throttled notice.

`considerAutomaticCompaction` does all the remaining deciding, unchanged — the threshold, `autoCompactionReady`, the open-continuation fence, the one-filing guard, the Astra/Auto-Off refusals. The only thing that moves is that "is this chat working" now has a third answer.

Safe to believe because the stretch is counted from **attributed** calls: the request-id join has tied each of them to this exact conversation. That is stronger evidence of work than a turn the page happens to be rendering, and it is the same evidence #280 already acts on when it reloads the page.

The log line says which of the three it was, so the reason a ticket exists stays readable.

## Two notes on the work itself

The regression sits at the end of `test/bridge.test.ts` on purpose, and says so: the blind-work notice is throttled once a quarter hour **across every chat**, so any test driving a blind stretch consumes that slot for whatever runs next. Placed earlier, it silenced the notice the `unattributed activity recovery` test asserts on — passing alone, failing in the file.

Chasing that also turned up a real defect in #280's own throttle, fixed there in `47662f4`: `now - lastBlindWorkNoticeAt < WINDOW` is satisfied by every negative difference, so a stamp that ends up in the future silences the notice until real time catches up — hours, after a sleep/wake or an NTP correction, about a page calling tools every few seconds.

Full suite on this branch: **4907 passed**, typecheck and privacy clean. Without the change the regression fails with no ticket, exactly as in the field.


totec448-spec:
An oversized current chat can already request compaction from its exact, unexpired MCP activity grant even when the page reports no running turn. The implementation rechecks that grant after storage reads and preserves completion, Stop, worker and departure guards. The stacked blind-page reload subsystem was intentionally not incorporated.

Reviewed head: `a286779f6fed`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #298, opened 2026-09-18
Retire the input a session left behind, so auto-continue can file again

Maximapple:
Auto-continue stops working for a session, permanently and silently, and the state that causes it is one row in `session-input.json`.

## What happens

`fileRecoveryInput` refuses to file while any non-terminal row for the session exists:

```ts
if (current.some(row => row.sessionId === sessionId && !terminal(row))) return false;
// terminal = ['sent', 'cancelled', 'failed']
```

That refusal is right — nothing may overtake authored input — but it assumes the row will resolve. A row in `browser` state addressed at a chat the session has **left** never will: Compact & Resume moves the session, and every guard on the way down checks the conversation, so the browser can neither deliver it nor fail it.

## Measured

From one install today, straight out of the durable store:

```
id              : 0f230404-68ad-43ee-a3c2-3e728289a51e
createdAt       : 09-18 09:52:04
state           : browser
sendAuthorizedAt: set
sessionId       : 2026-09-17-a2a83e9a      ← still the live session
conversationId  : 6aacf6ae-…               ← left at 17:19, two handoffs ago
```

Still there at 19:00, nine hours later. Across those nine hours the app filed **no further auto-continue for that session at all**.

That mattered at 17:37, when ChatGPT ended the prime's turn with a transport failure and never accepted another. The chat went blind, the workers finished their assignments and slept, the run parked — with the plan at `3 completed / 1 in_progress / 1 pending`. The person had to notice and type into the chat by hand. That is precisely the situation `recoveryMessage()` exists to handle, and it could not be filed. Nothing in the log said why.

`revokeSilenceInputs` does not reach these: it only strips claims from rows that were never authorized to send, and this one had been.

## The change

`armResumedChat` already knows the session has moved, so it retires what the old chat was still holding. Only rows whose address is provably stale — a row naming the current conversation is left alone — and the pass is idempotent, returning how many it retired so the log can say so when it is not zero.

Nothing about the refusal rule changes. A session with a live, deliverable row still files nothing, which is the point of it.

The regression pins the exact shape: an authorized `browser` row is untouched while it names the current chat, and retired once the session has moved; a second pass finds nothing.

Full suite: **4985 passed**, typecheck and privacy clean. Without the change the regression fails.


Repcoders:
+rep

Maximapple:
Third commit: the retirement now cancels **without** an error.

The renderer turns a cancelled row carrying an error into a visible notice, so every retirement put a failed "Carry on until everything requested is finished…" message into the chat with "the session moved to another chat before this could be delivered" underneath it. A user hit that within hours of the first commit landing on their machine, and they were right about it: they never wrote that message, it was never delivered, and there is nothing for them to act on.

The count stays in the log line each caller writes, which is where a housekeeping fact belongs. Both regressions now assert the absence of the error, so it cannot return as a notice unnoticed.

Full suite: **4986 passed**.

Maximapple:
Fourth commit, and the one that closes the shape rather than the instance.

Retiring rows a session has moved past covers the case where a handoff makes the address stale. It does not cover a session that never moves chats — and that wedges identically. An authorized browser row has no exit but its receipt: not a reload (`advanceRecoveryInput` matches only `sendAuthorizedAt === undefined`), not a timeout (the reclaim at the top of `listInputs` excludes authorized rows on purpose), nothing. Lose the receipt once and that session never files another automatic Continue.

Losing it is not exotic. Every recovery reload tears the page out from under a send in flight, and one machine saw twenty reloads in a day.

The send半 of that rule is right and stays: never resent, never cancelled on a clock, because the text may already be in the chat. What expires after fifteen minutes is only the row's **veto over new tickets** — past the forty seconds a bootstrap needs for its conversation id, past any receipt still in flight. A queued or unauthorized row keeps its veto for ever, because that is pending work.

Full suite: **4987 passed**. Without the change the regression fails on the boundary.

totec448-spec:
The demonstrated departed-chat Continue blockage is fixed through the existing outbox owner. Only an obsolete generated Continue with a proven session rebind is retired; authored queue entries and original late-receipt evidence remain intact. The blanket cancellation and fifteen-minute uncertain-send veto expiry are intentionally excluded. See docs/worklog-2026-09-19-pr-review.md and CONTRIBUTORS.md.

Reviewed head: `955ddb7b547d`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #299, opened 2026-09-18
Drop a worker failure from the wait caption once it is only history

Maximapple:
Reported by a user, in the words that made it obvious: *"das bleibt meist gleich, aktualisiert sich nie wirklich dynamisch."*

```
2 workers failed: Portable-Design, Repo-Validation
7 finished · 0 running · 2 failed
```

Unchanged for hours, across three Compact & Resume handoffs, naming two workers that had failed **the previous day**. Nothing was running. Nothing was being waited for.

## Why it stands there

The caption answers what the turn is waiting for, and it is produced whenever `workers.active > 0 || workers.failed > 0`. `active` falls back to zero on its own; `failed` never falls at all. So the moment a run has one failed worker, the wait caption is permanently occupied by it — and with `0 running` beside it, the line contradicts itself in the same breath.

The reading a user takes from that is not "two workers failed a while ago", it is "this display is stuck", which then costs the trust of every other caption next to it.

## The change

A failure counts towards the caption for fifteen minutes: long enough that it is still why the turn looks odd, short enough that it cannot outlive the work it belonged to.

A failure with **no** timestamp still counts. Absence of a time is not evidence that it is old, and dropping it would hide a fresh one — the wrong way to be wrong here.

Nothing else moves:

- the prime is told about each failure separately and permanently, by message, which is the durable record;
- `finished` and `total` are untouched, so the run's tally stays complete;
- a caption with something actually running is unchanged.

The regression fails the turn a worker into `failed`, asserts the caption names it, advances the clock past the window and asserts it is gone while `finished` is unchanged. Without the change it fails on the second half.

Full suite: **4984 passed**, typecheck and privacy clean.


Repcoders:
+rep

totec448-spec:
The stale worker wait caption is fixed at its presentation consumer: a worker-wait caption requires currently active workers. Historical failure counts remain complete in status/history, so the proposed fifteen-minute filter on shared worker counts was not taken. See docs/worklog-2026-09-19-pr-review.md; attribution is retained.

Reviewed head: `38a18dd66017`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #303, opened 2026-09-19
fix: preflight Travel Parent longrun creation

YahsanM:
Prevents singleton Longrun collision from burning TP child capacity. Adds remaining-capacity and renewal guidance while preserving the raw identity fence. Validation: 33/33 focused regressions; npm run typecheck; npm run build; git diff --check.


YahsanM:
Closing: this NEXORA custom hardening belongs on the custom 2.1.12 fork line and has already been merged there. Do not merge this divergent upstream comparison.

PR #305, opened 2026-09-19
Say when the browser never picked a command up at all

Maximapple:
Two different failures wear the same sentence, and the difference is the one fact that decides where to look next.

A command the browser **claimed** and could not finish is a page problem: it opened something, typed something, and the report never came. A command the browser **never claimed** is not — nothing was ever typed into any chat, so the page named in the message was never asked anything.

Both end as `the chat this app opened did not report back in time`.

## Measured

2026-09-19, a worker wake:

```
06:02:08  worker-11 is being woken in conversation 6aae1c96…
   …      (ninety seconds, no browser activity of any kind for this command)
06:03:38  could not wake worker-11 — the chat this app opened did not report back in time
06:03:38  gave up on revive:2ec1905a…:worker-11 — the chat this app opened did not report back in time
```

The wake channel had been authenticated since 05:03 and never disconnected. The worker's tab was still open — it was closed at 06:10, after this. And the same wake for the same worker had succeeded ten minutes earlier, in five seconds:

```
05:51:56  worker-11 is being woken in conversation 6aae1c96…
05:52:01  revival message reached worker-11; waiting for worker liveness
05:52:49  worker-11 revived from waking — conversation is still alive (turn)
```

I could not tell from the log that nothing had ever been typed. That sent me to the worker's page looking for a delivery that had never been attempted.

## The change

`claimedAt` already records it. The expiry reads it and says which of the two happened.

Nothing else moves — no deadline, no routing, and explicitly no retry: the comment above `COMMAND_DEADLINE_MS` is clear that a background retry loop was removed on purpose, and I am not putting one back on the strength of a single occurrence in a window where ChatGPT was also failing transports in neighbouring chats.

What this buys is that the next occurrence can be diagnosed from the log instead of reconstructed from tab-close timestamps.

The regression is the existing owner-null expiry test, which is exactly this case: it now pins the new sentence and pins that the page one is not used. Without the change it fails.

Full suite: **4984 passed**, typecheck and privacy clean.


Maximapple:
Second commit, and the measurement that the first one made possible.

The wake I reported above was not a one-off. It happened again three quarters of an hour later, so I timed every wake on that machine for the day:

| worker | seconds | outcome |
|---|---|---|
| worker-10 | 4.7 | delivered |
| worker-11 | 5.1 | delivered |
| worker-12 | 62.9 | delivered |
| worker-13 | 80.8 | delivered |
| worker-11 | **90.0** | **expired** |
| worker-10 | 6.7 | delivered |
| worker-12 | 15.5 | delivered |
| worker-13 | 23.8 | delivered |
| worker-11 | 25.6 | delivered |
| worker-10 | 79.6 | delivered |
| worker-11 | 36.8 | delivered |
| worker-13 | **90.0** | **expired** |

Ten delivered between 4.7 and 80.8 seconds, median 25.6. Two expired, both single wakes, both exactly at the ninety-second deadline. **The slowest successful delivery had nine seconds to spare.**

A deadline that close to the observed maximum is not a deadline. And a wake is not a generic command: the browser cannot start until its next `/status` pass, whose alarm has a thirty-second floor, then it focuses or reopens a tab and types, and several wakes from one prime message are delivered one at a time — visible above as 62.9 and 80.8 in the first batch of three.

The constant's own comment already records being raised from thirty seconds for this reason. Ninety is the same mistake with a larger number.

Three minutes clears the measured maximum twice over and covers a serialized batch of four at the alarm floor. It buys only patience: the wake is not retried either way, so this decides nothing except how long the app keeps believing one that is still on its way. Each false expiry costs a worker its whole task.

The regression pins the relation, not the number. Full suite: **4985 passed**.

totec448-spec:
Both timer and maintenance expiry now use the same claimed/unclaimed diagnostic. A wake has one absolute three-minute attempt; redeeming it does not renew that deadline or authorize a replay. All three commits in this reviewed head are covered by the adaptation documented in docs/worklog-2026-09-19-pr-review.md.

Reviewed head: `84f212e1f9ca`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #308, opened 2026-09-19
fix(windows): harden foreground activation

### LINKED ISSUE #306: Windows: foreground activation can fail when target thread is not attached
## Problem
On Windows, a visible background target window can fail foreground activation even when the caller and target are in the same session and integrity level. The helper currently attaches only the helper thread to the current foreground thread before calling `SetForegroundWindow`.

A real-device reproduction showed activation remaining on the previous foreground window and the connector correctly failing closed with `FOCUS_FAILED`.

## Root cause
The failing case required the target window thread to participate in the temporary input-queue attachment. Attaching only helper↔foreground was insufficient.

## Proposed change
Temporarily attach the distinct thread pairs helper↔foreground, helper↔target, and foreground↔target; call `SetForegroundWindow`; then detach only successfully attached pairs in reverse order.

This keeps the existing postcondition checks and does not add registry changes, synthetic ALT input, `SwitchToThisWindow`, or any relaxation of input-target validation.

## Validation
- Deterministic focus regression test updated for the three-pair attachment lifecycle.
- Windows focused computer tests: 22/22 passed.
- Typecheck passed.
- `git diff --check` passed.
- Real-device before/after: the installed behavior reproduced `FOCUS_FAILED`; the isolated patched focus routine successfully activated the same target, and the reverse direction also succeeded.
- Full `npm run verify` has one unrelated environment-sensitive `test/mcp.test.ts` failure caused by localized PowerShell parser-error text; the focus-related suites pass.

I have a narrow patch ready and can open it as a PR linked to this issue.

okayamajwcc-coder:
## What changed
Fixes #306.

### Root cause
Windows foreground activation could fail when the helper attached its input queue only to the current foreground thread. In the reproduced case, successful activation required the target window thread to participate in the temporary attachment as well.

### Smallest behavior change
`Focus()` now obtains the target thread and temporarily attaches the distinct pairs helper↔foreground, helper↔target, and foreground↔target. Only successful attachments are recorded, and they are detached in reverse order in `finally`.

The existing foreground postcondition remains authoritative. This does not use synthetic ALT input, registry changes, `SwitchToThisWindow`, or weaken input ownership/focus checks.

## Validation
- [ ] `npm run verify` passes. One unrelated environment-sensitive `test/mcp.test.ts` assertion fails because localized PowerShell emits different parser-error text; focus-related tests pass.
- [x] Packaging/runtime-relevant behavior was exercised with the isolated helper logic on Windows.

Focused Windows suites: 22/22 passed. Typecheck and `git diff --check` passed. On the same real machine, the installed behavior reproduced `FOCUS_FAILED`; the patched focus routine then activated that target successfully, and the reverse activation direction also succeeded.


totec448-spec:
The target-thread Windows focus repair is incorporated, with successful thread joins detached in reverse order even if a later attachment throws. Existing actual-foreground checks remain authoritative. The fresh C# regression compiled and passed against the current method with substituted native boundaries. CONTRIBUTORS.md retains Masatoshi Shisaka’s credit.

Reviewed head: `ed7bb2525de9`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #314, opened 2026-09-19
Reopen the chats this app still needs a page for

Maximapple:
Three ways a chat this app is working with ends up with no page, and is not given one back. The first two are halves of what a wake assumes and never checks; the third is the one that had been refusing reopens for the life of a chat.

## What it looks like when it does not

A wake is text this app is trying to type into one exact conversation. If no tab holds that conversation, the command is never claimed, nothing is typed anywhere, and three minutes later the worker is reported as failed to a prime that has been waiting the whole time.

Measured on one machine on 2026-09-19:

| worker | tab went | woken | outcome |
| --- | --- | --- | --- |
| worker-11 | 07:11:34, declined as "sleeping, not working" | 07:16:37 | unclaimed to the deadline |
| worker-11 | 5.3 s **after** the wake, declined as "waking, not working" | 07:16:37 | unclaimed to the deadline |
| worker-13 | 07:10:00 | 07:16:37 | unclaimed to the deadline |

Every wake whose tab survived was delivered, the fastest in 1.2 seconds. Both conversations still existed and were reachable; only their tabs were gone.

## The two gaps

**A slot that is `waking` was not reopened.** `queueMissingTab` reopens for `detached` only, and `waking` is the one other state that still owes its page something — the text is already queued for it. Narrow on purpose: only while a revival is actually pending for that exact conversation, so a slot left `waking` by something already given up on does not get a tab for work nobody is waiting on.

**A wake into a chat with no tab asked for nothing.** A sleeping worker's closed tab is correctly left closed. The other end of that decision was missing: the wake that arrives later is what makes that chat owed a page again.

## Why the reopen gets its own episode

A no-tab repair the browser has already carried out stays in flight until ordinary page activity retires it, and a chat whose tab never came back never produces any. Keying the wake's reopen on the command means the recovery a previous close already performed cannot stand in for it. Everything else is refused where it already was: a blocked chat, a stopped one, a chat with no session, and recovery the user turned off.

## The third: a slot that outlived its run

A run that ends is kept as dormant history, and every agent in it keeps the state it had at the end — a prime stays `active`, a worker stays `sleeping`. `agentInfoForOwnedConversation` answers out of that parked record when no live run owns the chat. That is the right answer to "was this ever a worker chat", which is what the compaction and Goal fences ask, and the wrong answer to "is a slot holding this chat now".

So a chat that had once hosted a swarm could never be given its tab back, for the rest of its life. Fifteen refusals over two days here, every one on a former prime, the last an hour after the app logged "restored no active run" at startup — while that chat's tool calls went on being filed under Unattributed activity, because nothing could say where the work was happening.

The missing-tab decision now reads every fact about a live slot from a live lookup, so a parked record neither bars the reopen nor grants the worker exemption. A chat with no live slot is judged the way any ordinary chat is: it must have called a tool, and something must be running in it.

## Tests

Three regressions, each of which fails on `main` with its production change reverted. The first draft of the waking one passed through neither branch — a chat this app has no session for never reaches the slot check at all — so the worker in it does a real turn before the wake, which is also what a woken worker always has behind it.

`bridge` + `agents` 661 passed, `session-input` + `mcp` 353 passed, typecheck and privacy clean.


totec448-spec:
The three demonstrated recovery cases are now fixed: parked prime history no longer vetoes a later ordinary chat turn, a waking worker can recover its missing tab, and a newly requested wake can recover the sleeping worker's absent chat. The integration reuses the existing recovery owner and additionally rechecks exact command ownership, undelivered messages, page return, manual closure, settings and bridge lifecycle after storage reads. Seven held-read negative cases pass alongside the original reproductions. Your contribution is credited in CONTRIBUTORS.md and the commit. Published in #316 (https://github.com/totec448-spec/chat-on-steroids/pull/316), merge commit `962d040b423f`. This is a source integration for the next release; it does not install or publish a new application version.

Reviewed head: `26053070de0b`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #315, opened 2026-09-19
Say what the marked replacement page named, once

Maximapple:
When a handoff ends `dispatched but never confirmed`, the question is whether the brief arrived at all — and the log could not say, so the search went looking at the send instead. It cost a day here in exactly that direction.

The replacement page reads the marker out of its own transcript and names the conversation holding it. That is the one fact this app cannot obtain any other way, and it was being answered silently. Now it says so once, with its outcome: bound and committing, refused because another message already holds it, or naming a continuation that no longer exists.

## Why the second commit exists

The marker stays in the replacement chat's transcript for good, so every later load of that page reads it again and names the same finished handoff. Measured here over four days: 62 such lines across 8 chats, 17 for a single chat, each saying the same thing about a continuation that had committed hours earlier.

Nothing is wrong when that happens — the 409 is the correct answer every time. But the first paragraph of this diagnostic promises "at most one line per replacement chat", and a log that repeats a settled fact is how a line that matters goes unread. That is not hypothetical: a monitor of mine went quiet against exactly this kind of noise the same week.

Keyed by chat and token, so a genuinely different handoff into the same chat still says its piece, and bounded like the other told-once sets beside it.

`bridge` 495 passed, typecheck and privacy clean. Diagnostic only — no behaviour outside the log changes.


totec448-spec:
Marked replacement outcomes are now logged with bounded deduplication. The adaptation records actual unknown/conflicting/retryable/rejected/committed outcomes, with commitment logged after settlement rather than before a possibly failing commit. Repeated transcript readback does not repeat the same outcome; the existing HTTP and continuation behavior is unchanged. Your contribution is credited in CONTRIBUTORS.md and the commit. Published in #316 (https://github.com/totec448-spec/chat-on-steroids/pull/316), merge commit `962d040b423f`. This is a source integration for the next release; it does not install or publish a new application version.

Reviewed head: `5ed455c50be9`. Compared with published main `c5ab88714d3b` (2.1.14).

PR #318, opened 2026-09-19
possible fix for the new a/b testing

ehkogh:
## What changed

All this does is make a half-fix for the new A/B testing, would call this incomplete in its entirely and NOT ready to be merged fully. I made it as a temporary fix while a more permanent fix would be available to use.

## Validation

- [x] `npm run verify` passes.


ehkogh:
fixes all into #319 so closed

PR #324, opened 2026-09-20
Do not make quit wait behind a connect that never returns

Maximapple:
Final shutdown queues its disconnect on the lifecycle queue, behind whatever is already running there. Nothing bounds a connect: `getSecret` waits on the OS keychain, which can put a prompt in front of it, and `startTunnel` waits on a remote handshake. When one of those never comes back, the queued disconnect never runs at all, and the whole admission phase sits out its 40s budget before being abandoned — with the local endpoint still listening, because the escape hatch above it only stops an endpoint that is already draining, never the live one a connect has just started.

It looks like this in the log — a shutdown that reaches `bridge stopped` and then goes quiet:

```
shutdown admission/drain starting
bridge stopped
shutdown admission/drain did not finish within 40000ms; continuing to quit
```

No `server stopping`, no `disconnected`: `disconnectImpl` never ran.

Two narrow changes: the immediate stop covers the live endpoint, and the queued teardown is raced against the same teardown run unqueued after a short grace. Running it twice is safe — it only acts on what is still live — and `shutdownRequested` keeps the connect it overtakes from starting anything once its own await finally returns.

The new test parks the keychain lookup forever, which is what actually happened here; the suite only had the case where it resumes. It fails without the fix. Measured on a packaged macOS build: quitting during a stalled connect went from 40s-and-abandoned to a clean drain.


Maximapple:
Force-pushed a corrected version of this branch. The original commit had a real defect in it, and it is the kind that only shows up under load, so it is worth saying plainly rather than burying it in a diff.

The unqueued teardown stepped around a drain that was already running instead of joining it. `disconnectImpl` takes custody of `endpoint` before it drains — the server moves to `drainingEndpoint` and `endpoint` becomes null — so a second call finds nothing to drain and returns almost immediately. `Promise.race` then accepted that as the phase answer. A quit taken while an MCP call was still writing would have ended phase 1 after the five-second grace, and the next phase kills exec processes and flushes the recorder and durable writers. The fix that was meant to stop a quit hanging would have cut a working drain short instead.

The branch now joins an in-flight drain (`stop()` shares one promise and its force deadline may only shorten, so joining cannot prolong anything), and carries a second test that drives a drain which is working and asserts the phase does not report itself done while it runs. That test fails against the version this PR had before.

Typecheck clean, `test/connection.test.ts` green (23), full suite green on the fork.

totec448-spec:
Reviewed at `bfee178667e5`.

The one-owner exit scheduling idea was already selectively incorporated and credited. Current main has the guarded exit timer and will-quit cleanup in window-lifecycle.ts. Closing this superseded branch without replacing the existing bounded exit grace.

PR #325, opened 2026-09-20
Escalate to a signal when neither exit ends the process

Maximapple:
On a packaged macOS build `app.exit(0)` returned instead of ending the process, and so did the `process.exit(0)` added to catch that: a probe written straight to a file — the logger is torn down by then — logged the line after each of them. Both run the C `exit`, which walks atexit handlers and static destructors, and one of those blocks while a native call is still outstanding.

What is left is a process with its window destroyed, its servers closed, thirty live threads parked in `mach_msg`, and the single-instance lock still held — so every later start of the app does nothing at all, silently. `sample` on it shows no exit frame anywhere.

A signal walks nothing, so it is the one ending that cannot be refused. It stays last: `app.exit(0)` is still asked first because it lets Electron tear down cleanly, and by the time the third step is reached every phase has flushed and there is nothing left to lose.

None of the three may be scheduled. A first attempt at a fallback sat behind a one-second timer that never fired — once Electron begins quitting it stops pumping Node event loop — so each step has to be the next statement, which a source-reading test pins.


totec448-spec:
Reviewed at `e88a210ecbf9`.

The existing exit owner retains a bounded grace after requesting Electron exit. A mocked synchronous return from app.exit does not demonstrate a wedged native process, and immediately forcing SIGKILL in finally removes that grace on normal exit too. Closing this change. The red Windows job was separately inspected: it failed live-plugin temporary-directory cleanup with EBUSY, not a shutdown assertion.

PR #341, opened 2026-09-20
File a compaction ticket when the run stops, not only while it runs

Maximapple:
Automatic compaction cannot save the workload it exists for, and I have three chats from one day that show it.

Every trigger requires the chat to be busy *now*. A run that calls tools without pause is exactly the run that cannot be compacted while it is busy: the page has to settle before it may ask for a brief, and settling needs a quiet moment a continuous run never offers.

Measured on 2026-09-20, same session chain, three chats:

- ticket filed at the threshold, declined by the page with `ChatGPT has not confirmed receiving the latest tool results`
- five raised reloads at two-minute intervals, the recording showing an unbroken sequence of `exec_command` / `write_stdin` / `read` calls straight through every one of them
- abandoned as `handoff_never_sent`
- the run then stopped at 471k context tokens, by which point ChatGPT was answering with `Message delivery timed out` — which is also the end of any brief, because ChatGPT writes it

All three had to be recovered by rebuilding the handoff out of the app own recording. That is not a feature.

The moment a run stops is the moment a handoff is easy: no turn to interrupt, nothing to settle. It was the one moment nothing could file a ticket. The silence sweep already wakes on exactly that — a grant the app was holding as it expires — so that wake-up now offers the chat to `considerAutomaticCompaction`, which rechecks every condition including the threshold. Idle old history is untouched: opening a finished conversation grants nothing and so starts nothing.

The stored refusal gives way for this caller alone. It is a verdict about the turn that would not take the handoff, held so a restart cannot refile it, and read as standing while no turn runs — right for a restart, wrong for the one caller that exists because the run ended. Bounded by its trigger: one offer per work episode, so a chat that refuses again refuses in seconds with a reason instead of ten minutes of reloads.

The test reproduces the sequence (over the line, a ticket filed and lost, then the run ending) and fails without either half.

`test/exec-shell-regressions.test.ts` has five failures on this branch base commit, before this change — shell-quoting cases, unrelated. Everything else is green.


Maximapple:
Force-pushed a correction. The first version of this branch was too narrow in two ways, and the machine it was written on showed both within the hour.

It hung the offer on `spent`, which means "this silence is finished with, nothing more is owed". A chat that goes quiet usually gets a silence repair in the same pass instead, and every branch that queues or awaits one leaves the loop before any `spent.push`. The chat this was written for then paused at 562k context tokens with repairs cycling every thirty seconds and never appeared in `spent` once — so the offer never came for exactly the chats that need it. The pass now also collects `stopped`: the plainer fact that the activity grant elapsed. A reload in flight is no argument against filing, since the fresh page reads a pending ticket back through `maybeResumePendingCompaction`.

It also wired the offer at one caller. `inspectSilentChats` has three, and the sweep that reached that chat was one of the other two — which is why the production run stayed silent while the test stayed green. The offer now lives inside the sweep.

The added test drives the branch that never reaches `spent`: recovery wants the tab, the chat goes quiet, the sweep queues its reload, and the ticket has to appear anyway. It fails without the `stopped` collection. `test/bridge.test.ts` is green (523).

captainhuke-dev:
i think it use handoff by text not md. I will full fill max context size.

totec448-spec:
Reviewed at `17acaa13c1bd`.

Elapsed work must not bypass the current-work requirement or the durable automatic-compaction refusal. That override can undo an explicit cancellation. The tracked local dependency link is unrelated. Closing this approach; the cancellation boundary itself is repaired in the focused integration.

PR #349, opened 2026-09-21
Auto-answer ChatGPT's Bussiness tool-approval card for this app's own connectors

aliceric27:
## What changed

Snipaste_2026-09-21_17-02-30

**Root cause.** On a ChatGPT Business workspace there is no "allow all tool calls" setting to
turn on. The account-level switch that would give a connector standing permission is not
available there, so what remains is the per-call approval card, plus `Allow → Always allow` —
which the control's own label scopes to *"for this conversation"*.

Every conversation therefore starts from zero, and this app creates conversations on the user's
behalf: Compact & Resume (§15) moves a live session to a new conversation `S: A → B` and fires
precisely when there is live work, and each worker (§16) owns another conversation. Attended,
that is a repeated click. Unattended it is a stall — an overnight Goal or Loop compacts, the new
conversation asks, nobody answers, and the run is idle until morning. The work was authorised;
there is simply no place on Business to say so once.

**The change.** The content script answers the card, for the connectors already in
`OUR_CONNECTORS` and no others.

- `extension/chatgpt-dom.js` finds the button by its **label**, not by `btn-primary`. Emphasis is
  not meaning: a redesign that emphasised Deny would otherwise have had this press Deny with full
  confidence. Emphasis now only breaks a tie between two affirmative labels, the dropdown half is
  excluded because it opens a menu, and an unnameable row yields nothing so the user clicks.
- `extension/content.js` presses it on card insertion after a 150 ms settle (the buttons mount in
  the same React commit and a click in that turn is swallowed) and again on the existing
  one-second observe tick. No new polling loop.
- `extension/popup.{html,js}` add **Auto-allow tool calls** beside Overwrite and Timestamps,
  stored as `autoApproveToolCalls`, on by default. **It is a plain UI switch the user can turn on
  or off at any time, with no reload, reinstall or restart**: the content script watches
  `chrome.storage.onChanged`, so a change applies immediately in every open ChatGPT tab. Turning
  it on also answers whatever card is already waiting; turning it off stops the next one. In
  effect this is the per-browser "allow all tool calls" switch Business does not offer, kept
  narrow and revocable.
  
  image

### Why the existing mechanisms do not cover it

`readOnlyHint` (`src/main/mcp/kernel.ts`) already removes the card for read-only tools, and that
line must not move — `exec_command` and file writes *are* write actions. This change touches no
annotation and widens no tool. The Setup reminder (`src/renderer/tool-approval.ts`) teaches
`Always allow` and correctly says it does not approve a call; on Business that is the only
remaining lever, and it cannot outlive the one conversation it was acted on.

### Safety

No provider safety decision, usage limit or account restriction is bypassed, and
`chrome.debugger`, the bridge, receipts and the paid-call path are untouched.

- The card's app line must match `OUR_CONNECTORS` exactly. Another connector's card is left
  alone; consent to this app is not consent to a mail or drive connector.
- Losing that app line to a redesign stops the automation rather than widening it.
- Bounded: 3 tries per card 2.5 s apart, and a rolling ceiling of 10 clicks a minute stops a card
  ChatGPT keeps redrawing after a failed call.
- **Freely revocable from the UI.** The popup switch is the only thing that grants this, one
  click takes it back, and "off" is honoured before the first card of the session — a browser
  that starts with it off never answers anything.

Not mirrored into the app's Settings yet: `BrowserPreferences` describes itself as browser-owned
*display* preferences, and this one takes an action. Four small edits if you would rather the app
own it too.

## Validation

- [ ] `npm run verify` passes.
- [ ] Packaging/runtime smoke w

| Check | Result |
| --- | --- |
| `npx vitest run test/tool-approval.test.ts` | 11 passed |
| `npx vitest run test/{tool-approval,content-script,extension,chatgpt-dom-input}.test.ts` | 1052 passed |
| `npm run typecheck` | clean |
| `npm run verify:privacy` | passed |

`test/tool-approval.test.ts` runs the shipped `chatgpt-dom.js` in jsdom against live-shaped
markup; the case it exists for is named in the test — *"still finds Allow when ChatGPT emphasises
Deny instead"* — with negative cases for the description's own button, the dropdown half, an
unnameable label, a disabled button, a hidden card and one drawn inside `.clf-stream`.

**Not run:** `npm run verify` in full — it includes `computer.test.ts`, whose real-desktop
foreground assertions need a desktop this was prepared without, so a failure there would be
environmental. No packaging smoke (`extension/*.js` ship unbundled). **Not verified in a live
signed-in ChatGPT page**; the card shape is from a 2026-09-21 capture, and the absence of an
"allow all tool calls" setting is reported from one Business workspace rather than measured
across plans or admin policies.


totec448-spec:
Reviewed at `b751a6534369`.

The approval preference is enabled by default, including after a preference-read failure. Connector names alone do not prove the installed app identity, and the action lacks an exact current owned-turn boundary. Plugin tools can also perform external side effects. Closing this version; an explicit opt-in design needs those identity and lifecycle guarantees before it can be considered.

Issue #336, opened 2026-09-20
Session stalled for hours while repeatedly attempting browser recovery, without resuming tool execution


captainhuke-dev:
### App version

2.1.14

### Operating system and version

Win10

### Architecture

x64 (Intel/AMD)

### Was the Chrome extension connected?

Yes, and its version matches the app

### ChatGPT plan

Free

### What happened

Chat On Steroids left my long-running task stuck in browser recovery for hours without successfully resuming execution.

### What happened

On September 20, 2026 (UTC+7):

- Around 09:23, the agent successfully installed an application release.
- At 09:31:40, the session recorded its last tool call.
- After that, the logs repeatedly showed:
  - "Trying to reload chat to recover an unresponsive open turn…"
  - "Trying to reload chat to recover an undelivered follow-up…"
- At 16:55:02, the session finally recorded:
  - "No visible progress for ten minutes. The app could not confirm that this turn finished."
  - `turn_end` with `outcome: "stalled"`
  - Detail: "no visible output and no progress for ten minutes"

That is over seven hours between the last recorded tool call and the terminal stalled event.

### Impact

The task was left unfinished. The application installation had succeeded, but the agent never completed its checkpoint and acceptance updates. The saved task plan still showed review/install work pending even though installation evidence existed.

This made it difficult to distinguish actual task progress from repeated recovery attempts. I had to inspect local logs and artifacts manually to discover where execution had stopped.

### Expected behavior

- Bound automatic recovery attempts with a clear timeout/retry limit.
- Surface an actionable stalled state promptly when recovery fails.
- Clearly distinguish "recovering the conversation" from "executing the task."
- Show the last successful tool execution time.
- Preserve the pending follow-up and explain whether it was delivered.
- On recovery, reconcile completed tool actions before repeating work.

### Environment / evidence

- Windows
- ChatGPT through Chat On Steroids Core
- Session ID: `2026-09-19-3afcc0e1`
- Session metadata:
  - `lastToolCallAt: 1789871500054`
  - `lastTurnEndAt: 1789898102471`
  - `lastTurnOutcome: "stalled"`
  - `activeTurnId: null`

I cannot determine from these logs whether the underlying failure originated in ChatGPT, the browser connection, or CoS itself. However, the recovery behavior failed to restore execution and took hours to reach a recorded stalled outcome.

An unattended task needs reliable failure reporting. Repeated reload attempts without restored execution are not meaningful progress.

### Anything else

```text

```

[...]

PR #358, opened 2026-09-21
End a repair the browser never claims, and remember that a chat has no page

Maximapple:
Re-offering an unclaimed repair is right: a claim can be missed, and the page that missed it is the one that needs the reload. It had no end, and the supply of repairs had none either.

Measured on one machine on 2026-09-20, after Chrome discarded a tab under memory pressure:

- **388 offers** of one repair between 12:46:54Z and 16:33:10Z, **exactly one ever confirmed**, still going at thirty-second intervals after three and a half hours. Other chats in the same log took 6, 3 and 2.
- Bounding the single repair was not enough. Over the next seven hours the chat went quiet, the sweep read that as silence, queued a fresh repair, and the new one arrived with its own budget: **four verdicts, 106 reload attempts, 600 offers**, every one into an empty room — and the same sentence written into the timeline four times where nobody was reading it.

So the offer gets a ceiling, and the verdict behind it is remembered instead of re-derived every five minutes. While it stands, recovery is declined for that chat. It is lifted by the only thing that can disprove it: the page polling for its own chat activity. Not a timer — nothing about elapsed time makes a discarded tab come back — and not durable, because a restart has no page to speak of either and should ask again.

What it buys is a sentence that could not be said before: *this chat has no page to reload, open its tab again and recovery resumes* — instead of a reload verdict about a page that was never there.

The test reads the offer forty times without ever claiming it, then hands the page-less chat a fresh failure and requires that it buys nothing, then lets the page come back and requires that recovery resumes. Against the unbounded version it is still being offered on the fortieth read. `test/bridge.test.ts` is green (528).


totec448-spec:
Reviewed at `2588e069e4ae`.

The status-poll count does not prove that the elected document is missing or that another opening is authorized. This creates a second page-absence decision alongside the existing browser owner and can conflict with intentional closure. The tracked local dependency link is also outside the fix. Closing this implementation; #336 stays open for the exact browser/command lifecycle evidence.

